PR #14, opened 2015-11-17
chore(fonts): rename + prefix CSS with dmn-icon-

### LINKED ISSUE #13: Prefix dmn-js internal fonts with dmn-
Currently the dmn-js internal icons have the class `icon-dmn icon-foobar`. We should change that to `dmn-icon-foobar` to avoid collisions with other fonts.

nikku:
Closes #13


nikku:
Merged via 7c9923aeec03b864de7f81aec46a8581e9d13254

PR #207, opened 2017-04-12
Add I18n support to dmn-js

iamfat:
Use the same mechanism of bpmn-js. People can customize translate function and load language file.

```javascript
import Modeler from 'dmn-js/lib/Modeler'
import translations from '@/i18n/zh.js'

var translateProvider = {
  translate: [ 'value', (template, replacements) => {
    replacements = replacements || {}
    // Translate
    template = translations[template] || template

    // Replace
    return template.replace(/{([^}]+)}/g, (_, key) => {
      return replacements[key] || '{' + key + '}'
    })
  }]
}

var modeler = new Modeler({
  additionalModules: [
    translateProvider
  ],
  table: {
    additionalModules: [
      translateProvider
    ]
  },
  disableDrdInteraction: true,
  container: '#dmn-container'
})

```

### DIFF
--- a/lib/table/features/date-edit/DateEdit.js
+++ b/lib/table/features/date-edit/DateEdit.js
@@ -21,7 +21,7 @@ var setValue = function(input, value) {
   }
 };
 
-function DateEdit(eventBus, simpleMode, elementRegistry, graphicsFactory, modeling, complexCell, selection) {
+function DateEdit(eventBus, simpleMode, elementRegistry, graphicsFactory, modeling, complexCell, selection, translate) {
   this._eventBus = eventBus;
   this._simpleMode = simpleMode;
   this._elementRegistry = elementRegistry;
@@ -29,6 +29,7 @@ function DateEdit(eventBus, simpleMode, elementRegistry, graphicsFactory, modeli
   this._modeling = modeling;
   this._complexCell = complexCell;
   this._selection = selection;
+  this._translate = translate;
 
   var refreshHandler = function() {
     if (this._simpleMode.isActive()) {
@@ -88,6 +89,7 @@ DateEdit.prototype.setupComplexCells = function() {
   var graphicsFactory = this._graphicsFactory;
   var elementRegistry = this._elementRegistry;
   var complexCell = this._complexCell;
+  var translate = this._translate;
 
   elementRegistry.forEach(function(element) {
     if (isDateCell(element)) {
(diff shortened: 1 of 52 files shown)


nikku:
CI is [currently failing](https://travis-ci.org/bpmn-io/dmn-js/builds/221586841) with your changes. Could you address the issues?

iamfat:
Yes, i'm looking at it..

iamfat:
@nikku I checked the code... it seems diagram-js changed after dmn-js's final commit. I've fixed one bug caused by issue#205 from diagram-js. But it seems that we still cannot pass the DecisionLabelUpdater test. I checkout the code to c4d539ca83cce78d75f1555ce4baa17c6130ded0 before I commit, it still failed on this test. Will that be caused by an unfinished feature?

Here is the test code
```javascript
    // given
    elementRegistry.get('dish-decision').businessObject.name = 'FOO BAR';
    // when
    eventBus.fire('view.switch');
    // then
    expect(elementRegistry.getGraphics('dish-decision').textContent).to.eql('FOO BAR');
```

iamfat:
Well.. solved.. it is still caused by diagram-js#205. Since diagram-js/lib/util/Text does not require parent element, I have to use svgAppend to append it after createText.

nikku:
Yes, incompatibility between dmn-js and the latest diagram-js may be the issue here.

lib/table/features/date-edit/DateEdit.js
```diff
@@ -113,10 +115,15 @@ DateEdit.prototype.setupComplexCells = function() {
       }
 
       var node;
+      translate;  // sorry, i have to declare it here in order to pass "no-unused-vars" test..
       if (element.column.type === 'dmn:InputClause') {
-        node = domify(require('./template-input.html'));
+        node = domify(require('./template-input.html').replace(/\{\{(.+?)\}\}/g, function(_, match) {
+          return eval(match);
```
nikku: Usage of `eval` is evil.

There should be another way to parse the templates, i.e. via a pattern and pipe it to translate.

lib/table/features/date-edit/template-input.html
```diff
@@ -1,18 +1,18 @@
 <div>
-  <h4>Edit Date Condition</h4>
+  <h4>{{ translate('Edit Date Condition') }}</h4>
   <select class="dateEdit-type-dropdown">
-    <option value="exact">Exactly</option>
-    <option value="before">Before</option>
-    <option value="after">After</option>
-    <option value="between">Between</option>
+    <option value="exact">{{ translate('Exactly') }}</option>
```
nikku: Please remove the `translate(` + `)` parts here and invoke translate from DateEdit.js.

nikku: (See comment above for details).

nikku:
PR looks good, except for one change:

Instead of using `eval` to evaluate `translate('foo')` calls in the template, why not parse the string from the template and pipe it to translate (outside the template). Using `eval` is considered evil, as many bad things may happen.

nikku:
Yes, still a compatibility issue. Need to upgrade dmn-js to work with updated diagram-js and rendering.

nikku:
Merged with some adjustments via https://github.com/bpmn-io/dmn-js/commit/bf3b3fba03e8ee81b63887036d75227f760d8541.

Issue #357, opened 2018-05-14
Data attributes in decision table viewer input headers and output cells


siffogh:
In Camunda BPM Cockpit, we use such data attributes to be able to modify input headers and output cells and in order to add values to them.

These data-attributes are already present in the Modeler but it would be necessary for us to have them in the Viewer.

nikku:
Would you be open to file a PR adding the required functionality?

* [ ] data-* attributes in viewer
* [ ] [tests](https://github.com/bpmn-io/dmn-js/blob/master/packages/dmn-js-decision-table/test/spec/features/decision-table-head/DecisionTableHeadSpec.js#L13) that verify the attributes exist

nikku:
Fixed via https://github.com/bpmn-io/dmn-js/pull/358.

nikku:
Released as `v4.4.0`.

PR #358, opened 2018-05-15
feat(decision-table): add data attribute to header input and output cells

siffogh:
This solves [issue #357 ](https://github.com/bpmn-io/dmn-js/issues/357).


siffogh:
I will add another commit to add data attributes to output cells.

PR #386, opened 2019-01-16
fix(shared/Manager): fire attach event in attachTo

barmac:
Viewers of each dmn-js library fire attach event
when they are attached to a container. The same is
the case regarding other libraries from the modelling
family, i.e. bpmn-js and cmmn-js. Thus, the change
is rather a fix than a feature since user expects
such event to be fired.


barmac:
Oh, so we don't require review here? @nikku

nikku:
We do :walking_man:...

nikku:
I fixed the repository config. :wind_face:

PR #392, opened 2019-02-19
fix(shared/Manager): fix viewers iteration on destroy()

ruvor:
It looks like someone has made a typo in viewers iteration: missed the iteration itself so individual viewers did not destroy themselves.

### DIFF
--- a/packages/dmn-js-shared/src/base/Manager.js
+++ b/packages/dmn-js-shared/src/base/Manager.js
@@ -275,7 +275,7 @@ export default class Manager {
   }
 
   destroy() {
-    Object.keys(this._viewers, (viewerId) => {
+    Object.keys(this._viewers).forEach((viewerId) => {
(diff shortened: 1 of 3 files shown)


nikku:
Would you like to contribute a [test case](https://github.com/bpmn-io/dmn-js/blob/master/packages/dmn-js-shared/test/spec/base/ManagerSpec.js#L278) that verifies the now fixed behavior?

ruvor:
Not sure if the test case suits enough but may be it'll do.

nikku:
Thanks for providing the test case, too!

I combined both commits and made sure we don't use internal API in the test case.

Merged via https://github.com/bpmn-io/dmn-js/commit/eae054dc8066244dd80a0c9b11eb24ba1cc892cf.

nikku:
Released as `v6.2.2`.

PR #395, opened 2019-02-28
feat(decision-table): show input and output label first in editors

### LINKED ISSUE #346: Change order of Input Expression/Name and Label
The Label can technically be omitted. But if you add an input expression "guests with children?", which looks fine on the first glance, (see this table: https://github.com/ingorichtsmeier/dmn-test-tool/blob/master/dmn-excel-tester/src/test/resources/exampleEvaluation/modeler12_1.dmn) you will get an error when you evaluate it: https://github.com/ingorichtsmeier/dmn-test-tool/blob/master/dmn-excel-tester/src/test/java/com/camunda/consulting/dmn_excel_tester/example/ExampleEvaluations.java#L169-L187

So, the Label is most important for business users and should be entered first. Then you could add an input expression for evaluation (or generate it on the fly as the simulator (https://camunda.com/de/dmn/simulator/) or the dmn-test-tool (https://github.com/ingorichtsmeier/dmn-test-tool) do.

Both, the Label and Input Expression should be visible all the time.

barmac:
Closes #346

Closes https://github.com/camunda/camunda-modeler/issues/1247

### DIFF
--- a/packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputEditor.js
+++ b/packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputEditor.js
@@ -92,6 +92,17 @@ export default class InputEditor extends Component {
     return (
       <div className="dms-container ref-input-editor">
 
+        <p className="dms-fill-row">
+          <label className="dms-label">Input Label</label>
+
+          <Input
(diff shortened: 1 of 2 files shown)


nikku:
This needs a minor release of dmn-js with the fix.

PR #398, opened 2019-04-09
fix(drd/DrdUpdater): do not crop connections straight after reconnection

barmac:
Requires https://github.com/bpmn-io/diagram-js/pull/317


barmac:
Sure, I'll do this.

barmac:
Done.

PR #416, opened 2019-10-24
fix(decision-table-head): correctly handle values erasing

barmac:
This fixes a problem of restoring the persisted value when
user erased the text. The unsaved `undefined` value was ignored
as falsy and then the persisted value was inserted into the input.

Related to https://github.com/camunda/camunda-modeler/issues/826

### DIFF
--- a/packages/dmn-js-decision-table/test/spec/features/decision-table-head/editor/DecisionTableHeadEditorSpec.js
+++ b/packages/dmn-js-decision-table/test/spec/features/decision-table-head/editor/DecisionTableHeadEditorSpec.js
@@ -1,4 +1,4 @@
-import { bootstrapModeler } from 'test/helper';
+import { bootstrapModeler, inject } from 'test/helper';
 
 import { query as domQuery } from 'min-dom';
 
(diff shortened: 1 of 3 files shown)


oguzeroglu:
Do we want to merge it to develop or master?

barmac:
Good catch. We want to merge it to `master`.

packages/dmn-js-decision-table/test/spec/features/decision-table-head/editor/DecisionTableHeadEditorSpec.js
```diff
@@ -14,64 +14,149 @@ import ModelingModule from 'src/features/modeling';
...
+        expect(input.value).to.eql('');
+      })
+    );
+
+
+    it('should correctly display erased output expression', inject(
+      function(debounceInput) {
+
+        // given
+        const inputCell = domQuery('th[data-col-id="input2"]', testContainer);
```
oguzeroglu: It seems like there's a copy-paste mistake here :P Could you convert the inputCell to outputCell, update the selector? We're testing input expression two times. So undoing changes in OutputCellContextMenu.js does not affect the test results.

barmac: Done.

oguzeroglu:
🎉

Issue #405, opened 2019-10-01
Default input expression language in header should not be FEEL


CarstenWickner:
__Is your feature request related to a problem? Please describe.__

By default, the input expression language in an input column header is "FEEL".
- As far as I understand, FEEL is meant to be used in input entries (i.e. table cells not table headers) for range and equality checks.
- In the DMN Engine, the default for input columns and output entries is/was JUEL. Only Input entries use FEEL by default.
- In older versions of dmn.js (some time ago) the default here was JUEL, which seems more meaningful.

<img width="576" alt="Bildschirmfoto 2019-10-17 um 13 14 40" src="https://user-images.githubusercontent.com/9433996/67004249-2bd17a80-f0e0-11e9-8d44-9c164139f04e.png">

<img width="522" alt="Bildschirmfoto 2019-10-17 um 13 19 43" src="https://user-images.githubusercontent.com/9433996/67004542-d5187080-f0e0-11e9-9990-ef41b013089f.png">

__Describe the solution you'd like__

The DMN Modeler's default input expression language (in input column headers) should match with the respective default in the DMN Engine.
Making this configurable (as per #19) is one option (especially since this is configurable on the DMN Engine). Changing the current hard-coding from FEEL to JUEL is another.

Change the default expression language for input column headers from _FEEL_ to _JUEL_ (see screenshots above)

* [ ] Change to JUEL in short hint 
* [ ] Set JUEL as the first option in script editor

__Additional Context__

* [Camunda Docs - Default Expression Languages](https://docs.camunda.org/manual/7.11/user-guide/dmn-engine/expressions-and-scripts/#default-expression-languages)

pinussilvestrus:
Hi Carsten!

Thanks for reporting, we will clarify this topic internally and come back to you whether we want to follow it this way in the future. 👍🏻 Stay tuned.

pinussilvestrus:
Hi @CarstenWickner, 

we decided that it absolutely makes sense to align the default expression language with the engine. I'll update the issue description and we will tackle this soon.

CarstenWickner:
Hi @pinussilvestrus,

Is there anything I can do to help? I started to make this all configurable (as per #19) to match the flexibility offered in the DMN Engine, but I'm struggling with your injection system. I don't know where to put the extra props so they would be passed through to where I need them, e.g. in the `features/decision-table-head` and `features/expression-language`.

Can you maybe give me some pointer for that pass-through of props?
[...]

Issue #19, opened 2015-11-24
I can set the default expression language for a table


gimbel:
May not be required

Migrated from: https://app.camunda.com/jira/browse/CAM-4696

nikku:
We'll not implement this any time soon unless we receive additional feedback.

nikku:
Implemented via https://github.com/bpmn-io/dmn-js/pull/441.

PR #418, opened 2019-10-29
Configurable expression languages

CarstenWickner:
The intention here is to introduce four new configuration options:
1. Setting the default Expression Language for Headers (and Output Cells) -- thereby also fixing the currently wrong default mentioned in #405.
2. Setting the default Expression Language for Input Cells -- just to be consistent with the above and avoid it being hard-coded in various places.
3. Setting the available Expression Language Options for Headers and Output Cells -- thereby allowing to hide unwanted options (e.g. if there is no Groovy or JRuby run-time where the decision table will be evaluated or FEEL just doesn't make sense) and add additional ones if desired.
4. Setting the available Expression Language Options for Input Cells -- same as point 3 but to enable different options (e.g. to only offer FEEL here).

The first two points are effectively #19.
The latter two points were already mentioned in my request from three years ago (for the non-react modeler):
> 7. Limit the available script languages for input expressions (to match the ones we actually provide on the classpath).
- https://forum.bpmn.io/t/customization-options-for-dmn-js-modeler/867

------

Related to #19 
Related to #405

### DIFF
--- a/packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputCellContextMenu.js
+++ b/packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputCellContextMenu.js
@@ -83,6 +83,8 @@ export default class InputCellContextMenu extends Component {
     return (
       <div className="context-menu-container input-edit">
         <InputEditor
+          defaultExpressionLanguage={ this.getValue('defaultExpressionLanguage') }
+          expressionLanguageOptions={ this.getValue('expressionLanguageOptions') }
           expressionLanguage={ this.getValue('expressionLanguage') }
           inputVariable={ this.getValue('inputVariable') }
           label={ this.getValue('label') }
--- a/packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputCellContextMenu.js
+++ b/packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputCellContextMenu.js
@@ -83,8 +83,8 @@ export default class InputCellContextMenu extends Component {
     return (
       <div className="context-menu-container input-edit">
         <InputEditor
-          defaultExpressionLanguage={ this.getValue('defaultExpressionLanguage') }
-          expressionLanguageOptions={ this.getValue('expressionLanguageOptions') }
+          defaultExpressionLanguage={ this.getValue('defaultInputExpressionLanguage') }
+          expressionLanguageOptions={ this.getValue('inputExpressionLanguageOptions') }
           expressionLanguage={ this.getValue('expressionLanguage') }
           inputVariable={ this.getValue('inputVariable') }
           label={ this.getValue('label') }
--- a/packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputEditor.js
+++ b/packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputEditor.js
@@ -20,7 +20,7 @@ export default class InputEditor extends Component {
       event.preventDefault();
       event.stopPropagation();
 
-      this.setExpressionLanguage('FEEL');
+      this.setExpressionLanguage('JUEL');
     };
 
     this.handleValue = (text) => {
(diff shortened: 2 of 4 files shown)


packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputCellContextMenu.js
```diff
@@ -83,6 +83,8 @@ export default class InputCellContextMenu extends Component {
     return (
       <div className="context-menu-container input-edit">
         <InputEditor
+          defaultExpressionLanguage={ this.getValue('defaultInputExpressionLanguage') }
+          expressionLanguageOptions={ this.getValue('inputExpressionLanguageOptions') }
```
CarstenWickner: This is one of the places where I struggle to understand how to pass those values through from the top-level modeler component.

packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputEditor.js
```diff
@@ -6,35 +6,25 @@ import ContentEditable from 'dmn-js-shared/lib/components/ContentEditable';
...
     this.setExpressionLanguage = (expressionLanguage) => {
       this.handleChange({ expressionLanguage });
     };
 
-    this.makeScript = (event) => {
-      event.preventDefault();
-      event.stopPropagation();
-
-      this.setExpressionLanguage('FEEL');
-    };
```
CarstenWickner: To simplify this a bit here, I've discarded the one-line input field and would show the multi-line editor every time.
As configuration user, I always find it annoying being forced to do that extra click for complex expressions and for simple ones I don't mind having a bit of empty space below either way.

Can of course revert this part, as it is not really necessary for making the expression languages configurable.

packages/dmn-js-decision-table/src/features/expression-language/ExpressionLanguage.js
```diff
@@ -44,8 +27,25 @@ export default class ExpressionLanguage {
...
         }
 
+        // this needs to come from somewhere
+        const configOptions = {};
+        const {
+          defaultInputExpressionLanguage = 'FEEL',
+          defaultHeaderAndOutputExpressisonLanguage = 'JUEL',
+          inputExpressionLanguageOptions = DEFAULT_INPUT_EXPRESSION_LANGUAGE_OPTIONS,
+          headerAndOutputExpressionLanguageOptions = DEFAULT_HEADER_AND_OUTPUT_EXPRESSION_LANGUAGE_OPTIONS
+        } = configOptions;
```
CarstenWickner: One more place I'd need a pointer as to where those need to be passed-through.

barmac:
Hi @CarstenWickner and thank you for your contribution. I modified a bit your first commit on [a new branch](https://github.com/bpmn-io/dmn-js/tree/418-JUEL_as_default_header_expression_language) so that the CI can build it and there are no misleading labels. I also adjusted the commit message to adhere to our [commit message guidlines](https://github.com/bpmn-io/dmn-js/blob/develop/docs/project/COMMIT_MESSAGES.md). After the weekend, we can continue with the configurability feature.

pinussilvestrus:
Thanks Maciej!

barmac:
@CarstenWickner I merged the change of the default expression language via https://github.com/bpmn-io/dmn-js/commit/6a80e3afad1c239215b4a6cf9490a821967ffab6.

barmac:
Regarding the configurability of the expression languages, this needs yet some more discussion. E.g. it might be worth it to actually configure one expression language for a whole column instead of separate cells. Please feel free to create an issue in which the feature might be discussed and then we can continue to work on this.

barmac:
Thanks for creating the issue. I'll get back to this PR later this or next week so that we can ship the actual feature.

barmac:
I'm closing this in favor of https://github.com/bpmn-io/dmn-js/pull/441

@CarstenWickner Thanks for your contribution!

nikku:
Merged via https://github.com/bpmn-io/dmn-js/pull/441. 

Thanks for contributing your initial version :+1:

PR #419, opened 2019-10-29
Bump table-js

barmac:


barmac:
It seems that table-js@6.0.1 might introduce a bug inside the dropdown. I'm closing this for now.

PR #441, opened 2019-11-27
Make expression languages configurable

### LINKED ISSUE #433: Configure Expression Language Options
As a user of an integrated Modeler instance, I want it to show all (and only) those expression language options that are available in the DMN-Engine being used (in the background), in order to minimise manual entry and reduce mistakes.

### Motivation
While the DMN-Engine allows for all kinds of expression languages being supported and even settings any default expression languages (for those expressions without a specific language selected), this flexibility is currently not available in the Modeler.
However, the Modeler might be exposed to users who are not aware of the Engine's configuration (and shouldn't have to be).

### Proposed Changes
Consequently I'd like to be able to configure the following things (in the order of priority):
1. **Show additional expression language options** that are specifically supported by the underlying DMN-Engine's configuration – e.g. "Scala", "Jython", or "FEEL-Scala" as requested in #333.
2. **Hide expression language options that are not supported**/covered by the underlying DMN-Engine's configuration – e.g. if there is no "Groovy" or "JRuby" runtime.
3. Be able to **override the default expression language** being applied – e.g. to use "FEEL-Scala" instead of "FEEL" in input cells (this becomes especially relevant if the standard "FEEL" is being hidden and the hard-coded default option becomes invalid.
4. **Hide expression language options that are not deemed meaningful in a particular context** – e.g. "FEEL" or "FEEL-Scala" for input (table header) expressions or output (cell) expressions would be a strange choice and the user shouldn't even consider those.

----

### References
Point 1 was already requested (for a particular option) in #333 and "solved" by introducing the ability to manually override the expression language (#322), but that means it would need to be overridden specifically everywhere which is not very nice.

Point 2 was already mentioned in my request from three years ago (for the non-react modeler):
> - Limit the available script languages for input expressions (to match the ones we actually provide on the classpath).
https://forum.bpmn.io/t/customization-options-for-dmn-js-modeler/867

Point 3 was requested already in #19 but closed because there was not enough vocal demand for it, but as mentioned above: when allowing to remove the hard-coded default options, making this configurable would avoid the need to introduce some intelligent fall-back logic.

This is supposed to be the basis for the discussion how far this configurability should go as requested by @barmac on the PR #418.

----

Related:
#19
#322
#333
#418

barmac:
This adds `ExpressionLanguages` module which reads the config
in order to provide available expression language options
as well as defaults for the editor and selected components.

Closes #433

### DIFF
--- a/packages/dmn-js-decision-table/test/spec/expression-language.dmn
+++ b/packages/dmn-js-decision-table/test/spec/expression-language.dmn
@@ -26,13 +26,13 @@
         <inputEntry id="inputEntry1" expressionLanguage="javascript">
           <text><![CDATA["bronze"]]></text>
         </inputEntry>
-        <inputEntry id="inputEntry2">
+        <inputEntry id="inputEntry2" expressionLanguage="feel">
           <text></text>
         </inputEntry>
         <outputEntry id="outputEntry1" expressionLanguage="javascript">
           <text><![CDATA["notok"]]></text>
         </outputEntry>
-        <outputEntry id="outputEntry2">
+        <outputEntry id="outputEntry2" expressionLanguage="juel">
           <text><![CDATA["work on your status first, as bronze you're not going to get anything"]]></text>
         </outputEntry>
       </rule>
--- a/packages/dmn-js-decision-table/test/spec/expression-language.dmn
+++ b/packages/dmn-js-decision-table/test/spec/expression-language.dmn
@@ -23,10 +23,10 @@
       <output id="output2" label="Reason" name="reason" typeRef="string" />
       <rule id="rule1">
         <description>Bronze is really not that good</description>
-        <inputEntry id="inputEntry1" expressionLanguage="javascript">
+        <inputEntry id="inputEntry1" expressionLanguage="jruby">
           <text><![CDATA["bronze"]]></text>
         </inputEntry>
-        <inputEntry id="inputEntry2" expressionLanguage="feel">
+        <inputEntry id="inputEntry2">
           <text></text>
         </inputEntry>
         <outputEntry id="outputEntry1" expressionLanguage="javascript">
--- a/packages/dmn-js-decision-table/test/spec/features/decision-table-head/editor/input/InputEditorCellSpec.js
+++ b/packages/dmn-js-decision-table/test/spec/features/decision-table-head/editor/input/InputEditorCellSpec.js
@@ -111,7 +111,7 @@ describe('decision-table-head/editor - input', function() {
 
       // then
       expect(inputBo.inputExpression.text).to.equal('foo\nbar');
-      expect(inputBo.inputExpression.expressionLanguage).to.equal('JUEL');
+      expect(inputBo.inputExpression.expressionLanguage).to.equal('juel');
     }));
 
 
@@ -126,7 +126,7 @@ describe('decision-table-head/editor - input', function() {
       triggerClick(makeScriptEl);
 
       // then
-      expect(inputBo.inputExpression.expressionLanguage).to.equal('JUEL');
+      expect(inputBo.inputExpression.expressionLanguage).to.equal('juel');
     }));
 
   });
--- a/packages/dmn-js-shared/src/features/expression-languages/ExpressionLanguages.js
+++ b/packages/dmn-js-shared/src/features/expression-languages/ExpressionLanguages.js
@@ -1,0 +1,133 @@
+import { assign } from 'min-dash';
+
+
+const EXPRESSION_LANGUAGE_OPTIONS = [{
+  label: 'FEEL',
+  value: 'feel'
+}, {
+  label: 'JUEL',
+  value: 'juel'
+}, {
+  label: 'JavaScript',
+  value: 'javascript'
+}, {
+  label: 'Groovy',
+  value: 'groovy'
+}, {
+  label: 'Python',
+  value: 'python'
+}, {
+  label: 'JRuby',
+  value: 'jruby'
+}];
+
+/**
+ * @typedef ExpressionLanguageDescriptor
+ * @property {string} value - value inserted into XML
+ * @property {string} label - human-readable label
+ */
+
+/**
+ * Provide options and defaults of expression languages via config.
+ *
+ * @example
+ *
+ * // there will be two languages available with FEEL as default
+ * const editor = new DmnJS({
+ *   expressionLanguages: {
+ *     options: [{
+ *       value: 'feel',
+ *       label: 'FEEL'
+ *     }, {
+ *       value: 'juel',
+ *       label: 'JUEL'
+ *     }],
+ *     defaults: {
+ *       editor: 'feel'
+ *     }
+ *   }
+ * })
+ */
+export default class ExpressionLanguages {
+  constructor(injector) {
+    this._injector = injector;
+
+    const config = injector.get('config.expressionLanguages') || {};
+
+    this._config = {
+      options: EXPRESSION_LANGUAGE_OPTIONS,
+      defaults: {
+        editor: 'juel',
+        inputCell: 'feel'
+      }
+    };
+
+    // first assign the list of languages as it might be required for the legacy defaults
+    if (config.options) {
+      this._config.options = config.options;
+    }
+
+    const legacyDefaults = this._getLegacyDefaults();
+
+    assign(this._config.defaults, legacyDefaults, config.defaults);
+  }
+
+  /**
+   * Get default expression language for a component or the editor if `componentName`
+   * is not provided.
+   *
+   * @param {string} [componentName]
+   * @returns {ExpressionLanguageDescriptor}
+   */
+  getDefault(componentName) {
+    const { defaults } = this._config;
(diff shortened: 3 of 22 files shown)


packages/dmn-js-decision-table/test/spec/features/decision-table-head/editor/input/InputEditorCellSpec.js
```diff
@@ -111,7 +111,7 @@ describe('decision-table-head/editor - input', function() {
 
       // then
       expect(inputBo.inputExpression.text).to.equal('foo\nbar');
-      expect(inputBo.inputExpression.expressionLanguage).to.equal('JUEL');
+      expect(inputBo.inputExpression.expressionLanguage).to.equal('juel');
```
nikku: Is this a breaking change or simply a change in the test case?

packages/dmn-js-shared/test/spec/features/expression-languages/ExpressionLanguagesSpec.js
```diff
@@ -0,0 +1,151 @@
...
+        defaultInputExpressionLanguage: 'jruby',
+        defaultOutputExpressionLanguage: 'javascript'
+      });
+
+      // when
+      const inputCellDefault = expressionLanguages.getDefault('inputCell');
+      const outputCellDefault = expressionLanguages.getDefault('outputCell');
+
+      // then
+      expect(inputCellDefault).to.have.property('value', 'jruby');
```
nikku: Do we have any other properties here? Otherwise a simple `expect(x).to.eql({ some: 'object', other: 'key' })` does the job, too.

packages/dmn-js-shared/test/spec/features/expression-languages/ExpressionLanguagesSpec.js
```diff
@@ -0,0 +1,151 @@
...
+          }
+        }
+      });
+
+      // when
+      const editorDefault = expressionLanguages.getDefault('editor');
+      const inputCellDefault = expressionLanguages.getDefault('inputCell');
+
+      // then
+      expect(editorDefault).to.have.property('value', 'javascript');
```
nikku: Do we have any other properties here? Otherwise a simple `expect(x).to.eql({ some: 'object', other: 'key' })` does the job, too.

packages/dmn-js-shared/test/spec/features/expression-languages/ExpressionLanguagesSpec.js
```diff
@@ -0,0 +1,151 @@
...
+    it('should set default input cell expression language to FEEL', () => {
+
+      // given
+      const expressionLanguages = createExpressionLanguages();
+
+      // when
+      const defaultLanguage = expressionLanguages.getDefault('inputCell');
+
+      // then
+      expect(defaultLanguage).to.have.property('value', 'feel');
```
nikku: Do we have any other properties here? Otherwise a simple `expect(x).to.eql({ some: 'object', other: 'key' })` does the job, too.

packages/dmn-js-shared/test/spec/features/expression-languages/ExpressionLanguagesSpec.js
```diff
@@ -0,0 +1,151 @@
...
+    it('should set default editor expression language to JUEL', () => {
+
+      // given
+      const expressionLanguages = createExpressionLanguages();
+
+      // when
+      const defaultLanguage = expressionLanguages.getDefault('editor');
+
+      // then
+      expect(defaultLanguage).to.have.property('value', 'juel');
```
nikku: Do we have any other properties here? Otherwise a simple `expect(x).to.eql({ some: 'object', other: 'key' })` does the job, too.

packages/dmn-js-decision-table/test/spec/expression-language.dmn
```diff
@@ -26,13 +26,13 @@
         <inputEntry id="inputEntry1" expressionLanguage="javascript">
           <text><![CDATA["bronze"]]></text>
         </inputEntry>
-        <inputEntry id="inputEntry2">
+        <inputEntry id="inputEntry2" expressionLanguage="feel">
```
nikku: Do we still test the default behavior (what if no expression language is set?)

packages/dmn-js-shared/test/spec/features/expression-languages/ExpressionLanguagesSpec.js
barmac: No. I will change this accordingly.

packages/dmn-js-decision-table/test/spec/expression-language.dmn
barmac: I've refactored the tests to make sure we do.

packages/dmn-js-shared/test/spec/features/expression-languages/ExpressionLanguagesSpec.js
barmac: Updated in the latest commit.

packages/dmn-js-decision-table/test/spec/expression-language.dmn
```diff
@@ -23,10 +23,10 @@
       <output id="output2" label="Reason" name="reason" typeRef="string" />
       <rule id="rule1">
         <description>Bronze is really not that good</description>
-        <inputEntry id="inputEntry1" expressionLanguage="javascript">
+        <inputEntry id="inputEntry1" expressionLanguage="jruby">
```
barmac: I changed this one because in one example `javascript` is set as default. Thus, it's no longer a non-default language.

nikku:
The fact that we do not have code coverage makes it quite hard to judge the actual test coverage.

I've opened https://github.com/bpmn-io/dmn-js/issues/442 to eventually have test coverage reporting inside our dmn-js delivery pipeline, too.

packages/dmn-js-shared/src/features/expression-languages/ExpressionLanguages.js
```diff
@@ -0,0 +1,133 @@
...
+export default class ExpressionLanguages {
+  constructor(injector) {
+    this._injector = injector;
+
+    const config = injector.get('config.expressionLanguages') || {};
+
+    this._config = {
+      options: EXPRESSION_LANGUAGE_OPTIONS,
+      defaults: {
+        editor: 'juel',
```
nikku: This indeed seems to be a breaking change: Previously the default expression language was `JUEL` in some places:

https://github.com/bpmn-io/dmn-js/pull/441/files#diff-7e4c370bbd7cff5b10aa86ee6d46b5f0L156
https://github.com/bpmn-io/dmn-js/pull/441/files#diff-39cd139761c02e1a63cafe3d4e248a49L11
https://github.com/bpmn-io/dmn-js/pull/441/files#diff-b6d2b22098e254991c1616651cb826c1L9

It was already `juel` in other places though:

https://github.com/bpmn-io/dmn-js/pull/441/files#diff-7e4c370bbd7cff5b10aa86ee6d46b5f0L137

Could you verify with the Camunda platform team if the capitalization matters at all when executing DMN? It looks like lower case is the way to go, at lease [looking at the DMN engine documentation](https://docs.camunda.org/manual/latest/reference/dmn11/decision-literal-expression/) and [scriptTask documentation](https://docs.camunda.org/manual/7.11/reference/bpmn20/tasks/script-task/). Or maybe the DMN standard documents can give hints whether it is `feel` or `FEEL`. The old behavior (mixing CAPITALIZED and lower cased languages) is clearly broken and deserves a cleanup. I'd just like to double check we're unifying into the right direction.

barmac: I'll clarify this with the Camunda platform team. If I understand DMN 1.1 Spec correctly, this should actually be an URI (cf. [page 41 of the spec](https://www.omg.org/spec/DMN/1.1/PDF#page=41)) :D

barmac: But for the Script Task it has to be [_mime-type format_](https://www.omg.org/spec/BPMN/2.0/PDF#page=195).

barmac:
I consulted the casing with the Camunda platform team and the results are as follows:
* the DMN Engine processes the expression languages disregarding the case, i.e. `feel` is treated as the same as `FEEL` or `FeEl`
* it's not documented
* the recommendation from the team is to use lowercased names.

ThorbenLindhauer:
Adding to this: I wasn't involved in the decision when this was built, but I'm pretty sure these acronym language names (e.g. `feel` instead of the full uri) were implemented for usability reasons when you are working directly with the xml. `feel` is just much easier to remember than an omg namespace uri. So it was purposefully not implemented 100% according to spec.

packages/dmn-js-shared/src/features/expression-languages/ExpressionLanguages.js
nikku: Thanks [for investigating](https://github.com/bpmn-io/dmn-js/pull/441#issuecomment-561194710).

PR #434, opened 2019-11-21
add i18n

leprechauns:
__Which issue does this PR address?_

Closes #

PR #446, opened 2019-12-11
Add i18n

barmac:
This Pull Request adds i18n and `translate` module to the decision table view.

Based on https://github.com/bpmn-io/dmn-js/pull/434

### DIFF
--- a/packages/dmn-js-decision-table/src/Viewer.js
+++ b/packages/dmn-js-decision-table/src/Viewer.js
@@ -9,6 +9,8 @@ import {
   remove as domRemove
 } from 'min-dom';
 
+import TranslateModule from 'diagram-js/lib/i18n/translate';
+
 import annotationsModule from './features/annotations';
 import coreModule from './core';
 import decisionTableHeadModule from './features/decision-table-head';
@@ -186,6 +188,7 @@ export default class Viewer extends Table {
     return [
       annotationsModule,
       coreModule,
+      TranslateModule,
       decisionTableHeadModule,
       decisionTablePropertiesModule,
       decisionRuleIndicesModule,
--- a/packages/dmn-js-decision-table/src/features/add-input-output/components/AddInput.js
+++ b/packages/dmn-js-decision-table/src/features/add-input-output/components/AddInput.js
@@ -10,6 +10,7 @@ export default class AddInput extends Component {
     this._eventBus = context.injector.get('eventBus');
 
     this._changeSupport = context.changeSupport;
+    this._translate = context.injector.get('translate');
   }
 
   onElementsChanged = () => {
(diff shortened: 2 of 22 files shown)


barmac:
There is one `TODO` in this PR, but all in all it seems to do its job. Still, I don't know why `npm run travis` fails. I have the same issue on `develop` as well as other branches.

nikku:
Travis must have changed something with their setup. All branches, even older commits that previously passed do now fail.

packages/dmn-js-decision-table/src/features/decision-rules/index.js
```diff
@@ -4,7 +4,8 @@ import DecisionRules from './DecisionRules';
 
 export default {
   __depends__: [
-    ExpressionLanguagesModule
+    ExpressionLanguagesModule,
+    'translate'
```
nikku: Does this do anything? We must depend on a module. Depending on a `string` does not yield any meaningful result.

packages/dmn-js-decision-table/src/features/decision-rules/editor.js
```diff
@@ -3,7 +3,7 @@ import DecisionRulesEditor from './DecisionRulesEditor';
 import Rules from './index';
 
 export default {
-  __depends__: [ DebounceInput, Rules ],
+  __depends__: [ DebounceInput, Rules, 'translate' ],
```
nikku: Does this do anything? I think we must depend on a module. Depending on a `string` does not yield any meaningful result.

packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/InputCellContextMenu.js
```diff
@@ -13,9 +13,7 @@ export default class InputCellContextMenu extends Component {
     super(props, context);
```
nikku: Unneeded change. Can we remove it?

packages/dmn-js-decision-table/src/features/decision-table-head/editor/index.js
```diff
@@ -17,7 +17,8 @@ export default {
     ContextMenu,
     DebounceInput,
     ExpressionLanguagesModule,
-    TypeRefEditing
+    TypeRefEditing,
```
nikku: Does this do anything? I think we must depend on a module. Depending on a `string` does not yield any meaningful result.

packages/dmn-js-decision-table/src/features/decision-table-head/index.js
```diff
@@ -1,6 +1,7 @@
 import DecisionTableHeadProvider from './DecisionTableHeadProvider';
 
 export default {
+  __depends__: [ 'translate' ],
```
nikku: Does this do anything? I think we must depend on a module. Depending on a `string` does not yield any meaningful result.

barmac:
> I'm not sure if depending on a string, i.e. `__depends__: [ 'translate' ]` does yield any meaningful result.
> 
> Getting rid of it and assuming that `translate` is provided per default is probably the better option here. We do the same in `bpmn-js`.
> 
> Other than that this one looks very good.

I'll remove that then.

Issue #450, opened 2020-01-10
Spike: A less flexible layout of the DRD (that could potentially be generated) is sufficient for our customers


pinussilvestrus:
__What should we do?__

Investigate the following hypothesis: "A less flexible layout of the DRD (that could potentially be generated) is sufficient for our customers", cf. [DMN Hypothesis Board](https://docs.google.com/spreadsheets/d/1ZI1GDUHJFP_RWLeIKFFxzFAlyJzH4Ozt2ApZnHIhVqQ/edit#gid=0).

__Why should we do it?__

A less flexible layout might allow building more understandable DRDs. How such a layout could look like and whether it offers benefits for our users, that is the task of this issue. The issue should track the progress of this Spike.

__Resources__

* [DMN Hypothesis Board](https://docs.google.com/spreadsheets/d/1ZI1GDUHJFP_RWLeIKFFxzFAlyJzH4Ozt2ApZnHIhVqQ/edit#gid=0)
* [Hypothesis 1 Summary Docs](https://docs.google.com/document/d/12IKMBsUJ6HfsptMYOLmwxZGaqKqB7KKDwPY14DRK0X8/edit?usp=sharing)
* [`dmn-prototypes`](https://github.com/pinussilvestrus/dmn-prototypes)

pinussilvestrus:
We did a little research on how to deal with large DRDs, to get ideas for the first hypotheses. Results can be displayed [here](https://docs.google.com/document/d/12IKMBsUJ6HfsptMYOLmwxZGaqKqB7KKDwPY14DRK0X8/edit?folder=1n7i_hFdisXFE8NqlWM1bAZi8f45lXqFn#heading=h.lqr5z1j7vgl2) (full) and [here](https://docs.google.com/presentation/d/1Hdt4B0pRVz2OoOojJbI0iGBIeY2paiyZig5Cz_Er5Mk/edit?usp=sharing) (summary).


We decided to do the first bunch of user tests with the practical modeling exercise and the horizontal static layout sketch.

pinussilvestrus:
**Exercise 1 - draw DRD layout**

[**Photo Documentary**](https://drive.google.com/drive/folders/11UjZHaST6r017_JP--s_KOyaGEFZDeoO?usp=sharing)

*User 1*

- Top-bottom is the preferred way, left-right not that often used
- Left-right looks more as BPMN
- DRDs are dependency trees
- Main rule on top of all others
- Modeling in stages
[...]

PR #454, opened 2020-01-24
Enable element distribution

nikku:
This PR enables proper element distribution in DRD diagrams.

Distribution is not blind to existing element aligments. If it finds certain axis it will group elements so that they are on the same axis (as seen in the screenshots below). So in this sense, it automatically generates levels as prototyped in https://github.com/bpmn-io/dmn-js/issues/450.

Also as seen in the screenshots below, we may want to invest some more time to properly fix connection layouts after distribution took place. I consider this a future improvement.

__Input__

image

__Distribute Horizontally__

image

__Distribute Vertically__

image

### DIFF
--- a/packages/dmn-js-drd/src/features/distribute-elements/DrdDistributeElements.js
+++ b/packages/dmn-js-drd/src/features/distribute-elements/DrdDistributeElements.js
@@ -16,9 +16,12 @@ export default function DrdDistributeElements(distributeElements) {
   distributeElements.registerFilter(function(elements) {
     return filter(elements, function(element) {
 
-      // TODO(nikku): add elements that cannot be distributed
       var cannotDistribute = isAny(element, [
-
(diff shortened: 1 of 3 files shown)


nikku:
> Maybe it's worth it to add a little test coverage for the filterung-out of certain elements? Like we have in bpmn-js, to ensure some elements are not distributed.

An integration test exists: https://github.com/bpmn-io/dmn-js/pull/454/commits/6628cefc371c81fa606fa995a19a06da7638050e#diff-6a3f1b2979317bc1c21c6522911d78adR33

It would fail with different exclusion patterns.

> The commit is named to be a fix. We usually don't direct fixing PRs to develop. I'd rather name it as a feat since we actively exclude elements from the distribution as an addition to the feature.

Technically it is a fix, as distribution was broken before explicitly excluding these elements.

PR #462, opened 2020-02-05
Correctly handle source element id change

barmac:
Related to https://github.com/camunda/camunda-modeler/issues/1679

Requires https://github.com/bpmn-io/dmn-js/pull/461


barmac:
@nikku Not sure why the CI status is _Pending_, as you can clearly see on Travis that the build passed.

packages/dmn-js-drd/src/features/modeling/behavior/index.js
```diff
@@ -1,14 +1,18 @@
...
 
 export default {
   __init__: [
     'createConnectionBehavior',
+    'idChangeBehavior',
     'replaceConnectionBehavior',
     'replaceElementBehavior'
   ],
   createConnectionBehavior: [ 'type', CreateConnectionBehavior ],
+  'idChangeBehavior': [ 'type', IdChangeBehavior ],
```
nikku: Why do we add the ticks `''` here?

PR #475, opened 2020-02-19
Indicate older DMN file incompatibility

### LINKED ISSUE #471: Capture opening attempts of older DMN standard diagrams and indicate issue with proper error message

People will update to dmn-js@8.0.0 as they always do. Only a few will read the changelog and understand the implications of the update. On diagram open, right now, we display the cryptic error message cannot parse diagram as `dmn:Definitions`.

To reduce our support effort we should capture opening attempts of older diagrams and indicate that with a proper (_TBD_) error message.

* [ ] Detect if diagram with older DMN standard version is being opened
* [ ] Output a proper error message that links to a help / learning resource (we may need to distinguish user facing errors from console.log statements. The later could be more descriptive).

Do nothing. This may cause big frustration for users and large amounts of support effort on our side.

__Additional context__

Detecting and indicating miss-use / common user errors is a good practice, seen in many libraries.

nikku:
Closes #471

### DIFF
--- a/packages/dmn-js-shared/src/base/Manager.js
+++ b/packages/dmn-js-shared/src/base/Manager.js
@@ -101,7 +101,7 @@ export default class Manager {
       this._setDefinitions(definitions);
 
       if (err) {
-        err = checkValidationError(err);
+        err = checkDMNCompatibilityError(err, xml) || checkValidationError(err) || err;
       }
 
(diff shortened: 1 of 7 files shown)


nikku:
I'm open for feedback if error and log statement are good enough :deciduous_tree:

PR #510, opened 2020-05-19
chore(hit-policy): improve viewer style

barmac:
After the hit policy PR was merged, I noticed that in the viewer it looks kinda odd. This PR is aimed to fix the problem.

Before:
image

After:
image

### DIFF
--- a/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
+++ b/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
@@ -242,6 +242,11 @@
   margin: 0 10px;
 }
 
+.dmn-decision-table-container .hit-policy-value::after {
+  content: '—';
+  margin-left: 5px;
+}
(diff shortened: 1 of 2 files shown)


volkergersabeck:
alternative maybe: have the hit policy label and the explanation text in the same font size, but separated by an m-dash (&mdash;)

pinussilvestrus:
or display the explanation in parentheses "(..)".

+1 for having the same font-size

barmac:
image

Let's have a look.

barmac:
I think I like the solution above more. What do you think?

nikku:
+1 for https://github.com/bpmn-io/dmn-js/pull/510#issuecomment-630877602.

barmac:
> +1 for [#510 (comment)](https://github.com/bpmn-io/dmn-js/pull/510#issuecomment-630877602).

It's on the branch now :)

PR #511, opened 2020-05-19
Rework decision table head

### LINKED ISSUE #498: Implement new ability to create columns to the right
__What should we do?__

Implement the new adding-column design

image

Note that,
* the adding circles are placed at the end the last input and output column
* centered vertically on the line
* the double border is getting thicker

__Why should we do it?__

We can get rid of the overloading global input/output header

Child of #494

### LINKED ISSUE #499: Implement new column header design
__What should we do?__

Implement the new design of input/output column headers

Cf. [Demo](https://dmn-prototypes.netlify.app/column-editing-final/)

image

image

* [ ] Every cell has a When/And/Then clause depending on its position
   * First Input Column: When
   * First Output Column: Then
   * n+2nd Input/Output column: And
* [ ] The input/output label is vertically centered and left-aligned
* [ ] The input/output label is a single line only
* [ ] The input/output label has a maximum width
* [ ] The input/output label got cropped when it exceeds the maximum width
* [ ] The input/output type OR allowed values are displayed in the bottom right corner

Things we can consider:
* Open current type editing modal when clicking on the bottom-right type label, as an intermediate step
* Keep the drag and drop functionality to change column order as it is
* Currently we show the expression if no label is filled. What shall we do if it is a multi-line script?

__Why should we do it?__

We can get rid of multiple cells for the headers

Child #494

barmac:
* Only one row for column labels.
* Input/output context menu extended with type ref.
* Moved "add column" button to the right edge of column group.
* Added input/output clauses.
* Moved drag'n'drop handle to top right corner.

Closes #498
Closes #499


oguzeroglu:
One feedback about this PR: It'd be really nice if the PR is easily reviewable -> Separate commits for each step.

philippfromme:
@oguzeroglu Since I'm more familiar with this project I can take over the review.

philippfromme:
One thing I noticed is that adding inputs and outputs has become harder because you only have a tiny button you can click. It used to be an entire cell.

philippfromme:
Also: There is no indication that the buttons that add columns are interactive other than a `cursor: pointer` which we don't want. We need another way of indicating that.

philippfromme:
Have we made any conscious decision about where the watermark should be after changing the table head? Right now it sits quite awkwardly next to the table:

image

barmac:
> Also: There is no indication that the buttons that add columns are interactive other than a `cursor: pointer` which we don't want. We need another way of indicating that.

Can you suggest any solution to the problem?

philippfromme:
> Can you suggest any solution to the problem?

I'd probably color the buttons to begin with so they stand out more. I tried a blue background, a white outline and centering them a bit better:

image

The outline color could change on hover.

barmac:
For comparison, this is the way we used to highlight the button:

Kapture 2020-05-26 at 9 20 26

barmac:
> Have we made any conscious decision about where the watermark should be after changing the table head? Right now it sits quite awkwardly next to the table:
> 
> image

Let's discuss it separately.

philippfromme:
> Let's discuss it separately.

In a separate issue you mean?

barmac:
Exactly. I think it's outside of the scope of this pull request.

barmac:
I think also that the rebranding stuff will touch the topic of logo position.

philippfromme:
Alright, so we definitely need to have another hover indication instead of the pointer.

Also: I noticed that the _Hit Policy_ label is the only text in the entire table that's bold. From a visual hierarchy point of view that makes little sense to me.

philippfromme:
image

Did we decide to move the _View DRD_ button and logo above the table? I know this wasn't considered in the prototype but we have to deal with it somehow. As I've already pointed out if you do not have the _View DRD_ button the logo ends up right of the table which doesn't seem to be something that we want.

barmac:
> image
> 
> Did we decide to move the _View DRD_ button and logo above the table? I know this wasn't considered in the prototype but we have to deal with it somehow. As I've already pointed out if you do not have the _View DRD_ button the logo ends up right of the table which doesn't seem to be something that we want.

This one slipped as we already disabled the logo in Camunda Modeler. In my opinion the logo should not float to the right as with fixed-width of the columns it might land outside of the area we want it to be, as can be seen on your screenshot.

philippfromme:
Another question: Have we thought about literal expressions? They also have a header. Have we deliberately decided to not apply the same changes?

image

pinussilvestrus:
> Have we thought about literal expressions

No, we have not discussed it right now. Maybe it makes sense to have the same changes there (remove id field etc.). But this have to go in another issue.

nikku:
We should consider aligning both in the next dmn-js release.

philippfromme:
Maybe we should have a separate issue for literal expressions then.

pinussilvestrus:
Done: https://github.com/bpmn-io/dmn-js/issues/515

Feel free to add additional work items on the list. Thanks for mentioning this one @philippfromme

barmac:
button

I just pushed changed hover indicator. Shipped with increased padding for the button and `cursor: pointer` removal.

barmac:
@philippfromme how does it feel now?

nikku:
Looks much better.

barmac:
@philippfromme Do I still need to change anything here?

philippfromme:
Not for now. I'll have another look at this.

philippfromme:
The clickable area is rectangular whereas the visible button is circular which I find unintuitive. I even partially obscures the text in the cell on the right.

image

My suggestion is something similar to what I already commented:

suggestion

I changed a couple of things:

* moved the button 1px to the right so it's more in the center
* smaller padding of 4px so the clickable area doesn't obscure the text in the cell on the right

image

* added a box-shadow around the button so is stands out a bit more and doesn't clash with the table borders

image

PR #531, opened 2020-07-02
chore(decision-table): update table-js to v6.1.0

barmac:
This is actually required for the new column editing feature.


oguzeroglu:
If CI works, I'll approve this 🐒

PR #544, opened 2020-07-09
chore(CHANGELOG): update to v9.0.0

barmac:


barmac:
Let's release it then.

PR #565, opened 2020-08-06
Context menu anywhere on rules

### LINKED ISSUE #555: Be able to remove row when clicking on leftmost column

DMN decision table columns can be removed by right-clicking in the column header.
The same is not possible for rows. They can only be removed by clicking into a respective cell. 

The experience should be consistent for rows and colums.

- [ ] Be able to remove a row via context-menu which is opened via right-clicking into the leftmost column.
- [ ] Be able to remove a column via context-menu which is opened via right-clicking into a cell.

Alternatively we could only allow adding / removing rows and columns via the leftmost and top row.

__Additional context__

Enhancement collected during DMN user training.

barmac:

Closes #555

__Definition of Done__

### DIFF
--- a/packages/dmn-js-decision-table/src/features/annotations/components/AnnotationCell.js
+++ b/packages/dmn-js-decision-table/src/features/annotations/components/AnnotationCell.js
@@ -12,7 +12,7 @@ export default function AnnotationCell(props) {
   } = row.businessObject;
 
   return (
-    <Cell className="annotation" elementId={ id + '__annotation' }>
+    <Cell className="annotation" elementId={ id }>
       { description || '-' }
     </Cell>
   );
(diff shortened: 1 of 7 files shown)


packages/dmn-js-decision-table/src/features/annotations/components/AnnotationCell.js
```diff
@@ -12,7 +12,7 @@ export default function AnnotationCell(props) {
   } = row.businessObject;
 
   return (
-    <Cell className="annotation" elementId={ id + '__annotation' }>
```
pinussilvestrus, line 15: Why we had this suffix in the first place?

pinussilvestrus: Kapture 2020-08-07 at 8 36 28

Seems like this is causing unwanted side effects. The annotation content will be displayed as description for some reason.

barmac: > Why we had this suffix in the first place?

That's a good question. It does not lead to any element of XML and is not used anywhere.

barmac: Fixed :)

PR #597, opened 2020-10-15
Allow decision name to take empty space

### LINKED ISSUE #579: Allocate more space for a decision name

[This issue was collected by PM in a user feedback / testing session].
Long Decision names are currently not entirely visible in the DMN decision table view. 
LongDecisionName

In the dmn decision table header, next to the Hit policy there is still plenty of empty space. This space shall be used when a long decision name is entered. For example:
* No max-width for the decision name
* Define min-width for the Hit Policy span
* Max-width of the table header should not be large than the table body

Alternatively one can move the Hit Policy span to the right side even if a decision name is short.

__Additional context__

barmac:
Closes #579

__Definition of Done__

### DIFF
--- a/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
+++ b/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
@@ -85,7 +85,7 @@
 }
 
 .dmn-decision-table-container .tjs-container {
-  display: grid;
+  display: inline-grid;
   grid-template-columns: min-content auto;
 
 }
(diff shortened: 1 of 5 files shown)


packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
```diff
@@ -258,10 +258,12 @@
 .dmn-decision-table-container .decision-table-name {
   position: relative;
   display: inline-block;
```
MaxTru, line 260: Since the parent has `display: flex`, I don't think we need this, do we? (aren't we mixing concepts then?)

MaxTru (approved):
Screenshot_20201016_105806
Works good! :+1: Only 1 comment for clarification

But do we still need

packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
barmac: Yeah, it makes sense. I've pushed a change for that plug a few `span -> div` changes as we use blocks.

PR #598, opened 2020-10-15
Correctly update examples

barmac:
This allows to keep `package-lock.json` files updated accordingly.

Related to https://github.com/bpmn-io/dmn-js-examples/issues/18

__Definition of Done__

### DIFF
--- a/tasks/stages/update-examples
+++ b/tasks/stages/update-examples
@@ -16,8 +16,13 @@ cd "$EXAMPLES_DIR"
 TOOLKIT_VERSION="${TRAVIS_TAG:1}"
 echo "Updating toolkit versions to $TOOLKIT_VERSION"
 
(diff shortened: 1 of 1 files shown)


nikku:
What was wrong about the old way of updating?

From what I understand this PR now force installs dmn-js, if it is used or not, does not matter.

nikku:
To fix bpmn-io/dmn-js-examples#18 we could update the package-lock after each "patch", too.

nikku:
Obviously one could argue that every project in dmn-js-examples should depend on dmn-js. In this case the PR is :rocket:.

barmac:
> What was wrong about the old way of updating?

It was impossible to run `npm ci `.

> Obviously one could argue that every project in dmn-js-examples should depend on dmn-js. In this case the PR is 🚀.

Yeah, that's what I'd expect.

nikku:
All right then.

nikku:
Alternative, as mentioned would have been to rebuild the package lock, which is a reasonable behavior after updating the examples, too.

PR #621, opened 2021-01-18
test: disable focus tests on Firefox

nikku:
These tests only work once in a while.

We cannot have a sporadically failing test suite.

Closes https://github.com/bpmn-io/dmn-js/issues/618


philippfromme (approved):
👏

PR #635, opened 2021-06-03
Allow to add output if dmn:input does not exist

barmac:
Previously the editor would fail to add an output if no input was given in the DMN table. This PR fixes the behavior.

### DIFF
--- a/packages/dmn-js-decision-table/src/features/editor-actions/DecisionTableEditorActions.js
+++ b/packages/dmn-js-decision-table/src/features/editor-actions/DecisionTableEditorActions.js
@@ -115,7 +115,8 @@ export default class DecisionTableEditorActions {
         const root = sheet.getRoot(),
               businessObject = root.businessObject;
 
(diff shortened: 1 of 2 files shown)


nikku:
Added a (maybe useful) PR description.

nikku:
Reminder to remove the unused PR template [which I did](https://github.com/bpmn-io/dmn-js/commit/1c012c3be070c205edf184f7b550c1f3d0c5e394).

pinussilvestrus (approved):
nice catch 👍 

interesting that we were prepared for it in the [`addInput` action](https://github.com/bpmn-io/dmn-js/blob/d6e74f6168fd421dfe4a47cd6952a0c4a9b18372/packages/dmn-js-decision-table/src/features/editor-actions/DecisionTableEditorActions.js#L65) but not the other way around.

barmac:
I don't get why codecov says it's not covered while it's covered o.O

Issue #514, opened 2020-05-26
Switch from callback to promises


nikku:
> **This is an epic issue which may be broken down into multiple tasks, if needed**

__What should we do?__

The core public APIs for dmn-js shall be migrated from Callbacks to Promises. This affects:

**Public APIs**
- [x] `DmnModdle#fromXML` (https://github.com/bpmn-io/dmn-moddle/issues/19)
- [x] `DmnModdle#toXML` (https://github.com/bpmn-io/dmn-moddle/issues/19)
- [x] `Manager#importXML`
- [x] `Manager#saveXML`
- [x] `Viewer#saveSVG`

**Internal APIs (e.g., communication between `Manager`, `table`, ...)**
- [x] `Manager#open` and `Viewer#open` (https://github.com/bpmn-io/dmn-js/pull/636)

Note: this leads to not supporting IE11 anymore, one would need to polyfill promises in this case.

Doing this shall follow the same approach as the [bpmn-js promisifaction in early 2020](https://github.com/bpmn-io/bpmn-js/issues/812). This approach set the following cornerstones:

* Don't use a wrapper but modify each API (e.g., public API shall inherently return a promise, the private API may still be callback-based). (See [RfC](https://github.com/bpmn-io/rfcs/tree/master/001-callbacks-to-async-apis))
* Ship major releases once this was implemented (for `DmnModdle` as well as for `dmn-js`)
* If a user still passes callbacks to dmn-js, then issue an error. But don't crash. Show a meaningful error message linking to "how to migrate from callback to promise APIs".
* For DmnModdle => don't offer callback API anymore (same approach as for BpmnModdle)

In particular we need to:

- [ ] Implement and release the change
- [ ] Update the documentation (incl. examples in other repos)
- [ ] Create a blog post for the major release (new deep link / or update the old article to also be meaningful for dmn-js)

__Why should we do it?__

* We should provide consistent developer experience across our libraries. 
* Promises nicely integrate with `async`/`await`

__Context__
[bpmn-js RfC](https://github.com/bpmn-io/rfcs/tree/master/001-callbacks-to-async-apis)
[bpmn-js promisifaction in early 2020](https://github.com/bpmn-io/bpmn-js/issues/812)

MaxTru:
Closed with https://github.com/bpmn-io/dmn-js/issues/641

nikku:
Do we plan to follow up with updates to our examples and a potential blog post?

MaxTru:
[...]

PR #636, opened 2021-06-08
Promisify open api

MaxTru:
related to #514

* This PR adds:
  * promisification of the existing open API
    * dmn-js-decision-table Viewer#open
    * dmn-js-drd Viewer#open
    * dmn-js-literal-expression Viewer#open
    * dmn-js-shared View#open
    * dmn-js-shared Manager#open
  * on success, #open will resolve to a {OpenResult} which contains a
    `warnings` property of type {array<string>}
  * on error, #open will reject to a {OpenError} which contains a
    a `warnings` property of type {array<String>}
* Tests still use `done`, this will be changed once the overall
  public APIs were promisified

BREAKING CHANGES:

* The (Manager#open | View#open | (dmn-js-decision-table |
  dmn-js-literal-expression | dmn-js-drd) Viewer#open) APIs now return a
  Promise. Note that passing callbacks to Manager#open is still
  possible, but deprecated. An error will be logged when doing so.
* Users are now expected to have Promises either by default or
  polyfilled.


packages/dmn-js-decision-table/src/Viewer.js
```diff
@@ -36,31 +36,62 @@ export default class Viewer extends Table {
...
-        // clear existing rendered diagram
-        this.clear();
+  /**
+   * Open diagram element.
+   *
+   * @param  {ModdleElement} decision
+   * @returns {Promise} Resolves with {OpenResult} when successful
+   * or rejects with {OpenError}
+   */
+  open(decision) {
```
pinussilvestrus, line 57: So we won't have this fallback wrapper for this (at least for a while), right? Will we have it for other APIs of dmn-js or do we completely follow the update route here?

https://github.com/bpmn-io/bpmn-js/blob/c3e0d6d16477bc2e9e37109a9ff1254b6695c7cc/lib/util/CompatibilityUtil.js#L13

MaxTru: Yes we will have it, but only for the key public APIs:

Screenshot_20210608_151935

pinussilvestrus: In bpmn-js, we also do it for other APIs, as [`open`](https://github.com/bpmn-io/bpmn-js/blob/develop/lib/BaseViewer.js#L279). The question is whether we want to have it the same for dmn-js.

MaxTru: thanks for highlighting. I will look into this.

MaxTru: I added the `wrapForCompatability` around `Manager#open`.

Reason:
* `Manager#open` is the public entry point when using dmn-js and wanting to open a View
  * We expect library users to use `dmn-js` as a whole. (e.g., loading the diagram using dmn-js#importXML, then switching view with `dmn-js#open` which goes via the `Manager`)
  * Although using specific views individually, this is (IMO) not a supported public API use-case (e.g., see that we don't tests these APIs stand-alone, also there are no examples around this)
  * Therefore I propose not to `wrapForCompatability` in the specific Viewers, since this would only blow-up the solution.

Happy to discuss

nikku: Makes absolute sense. :+1:

packages/dmn-js-drd/src/Viewer.js
```diff
@@ -224,30 +224,64 @@ Viewer.prototype._createContainer = function() {
...
+      }
+
+      // update definitions
+      self._definitions = definitions;
 
-  return done(err);
+      // perform graphical import
+      return importDRD(self, definitions, function(err, warnings) {
+        if (err) {
+          err.warnigns = warnings || [];
```
pinussilvestrus, line 264: little typo: `err.warnings`

I wonder why this never fails, though ...

packages/dmn-js-shared/src/base/Manager.js
```diff
@@ -316,13 +331,32 @@ export default class Manager {
...
    *
-   * @param  {ModdleElement}   element
-   * @param  {Function} [done]
+   * @param  {View} view
+   * @returns {Promise} Resolves with {OpenResult} when successful
+   * or rejects with {OpenError}
    */
-  open(view, done=noop) {
-    this._switchView(view, done);
+  open(view) {
```
pinussilvestrus, line 349: This looks a bit messy. Maybe we can make the wrapping earlier? Something like (didn't try it out)

```js
open = wrapForCompatibility((view) => {
  // ...
})
```

packages/dmn-js-shared/src/base/Manager.js
```diff
@@ -112,7 +112,7 @@ export default class Manager {
       }
 
       if (err || !options.open) {
-        this._emit('import.done', { error: err, warmings: parseWarnings });
+        this._emit('import.done', { error: err, warnings: parseWarnings });
```
barmac, line 124: Good catch! 🙀

packages/dmn-js-shared/test/spec/base/ManagerSpec.js
```diff
@@ -10,7 +10,8 @@ import { find } from 'min-dash';
 
 class TestViewer extends Manager {
 
-  constructor(viewProviders=[ DECISION_TABLE_VIEW, DRD_VIEW ], options={}) {
+  constructor(viewProviders=[ DECISION_TABLE_VIEW, DRD_VIEW ],
+      options={}) {
```
pinussilvestrus, line 14:
```suggestion
  constructor(
    viewProviders=[ DECISION_TABLE_VIEW, DRD_VIEW ],
    options={}
  ) {
```

pinussilvestrus: Maybe even cleaner

packages/dmn-js-shared/test/spec/base/ManagerSpec.js
```diff
@@ -297,17 +370,19 @@ describe('Manager', function() {
             manager._updateViews();
 
             // then
-            expect(viewsChangedSpy).to.have.been.called;
+            expect(viewsChangedSpy).to.not.have.been.called;
```
pinussilvestrus, line 373: Is this a behavior change related to the promisification? Or was it a bug in the existing test suite?

packages/dmn-js-shared/test/spec/base/ManagerSpec.js
```diff
@@ -497,18 +575,85 @@ describe('Manager', function() {
...
+              expect(error).to.exist;
+              expect(error).to.be.an.instanceof(Error);
+              expect(error.message).to.equal('foobar');
+
+              expect(error.warnings).to.exist;
+              expect(error.warnings).to.be.an.instanceof(Array);
+              expect(error.warnings).to.have.length(2);
+
+
+
```
pinussilvestrus, line 651:
```suggestion

```

pinussilvestrus: Lots of spaces :)

packages/dmn-js-shared/test/spec/base/ManagerSpec.js
```diff
@@ -22,6 +23,37 @@ class TestViewer extends Manager {
 
 }
 
+class ReturnWarningsView extends TestView {
```
pinussilvestrus, line 26: Can you quickly explain why we add these test views here? For better testing of view changes?

packages/dmn-js/test/spec/NavigatedViewerSpec.js
```diff
@@ -103,7 +113,12 @@ describe('NavigatedViewer', function() {
       // can open decisions
       expect(drdView.element.$instanceOf('dmn:Definitions')).to.be.true;
 
-      editor.open(drdView, done);
+      editor.open(drdView)
+        .then(
+          result => done(result.warnings[0]))
```
pinussilvestrus, line 118: Maybe add a note that we have to remove the `done` here as well. But we will probably remove all of them once the migration is done.

pinussilvestrus (commented):
Some comments regarding code style. Looks pretty good so far 👍

packages/dmn-js-shared/src/base/Manager.js
MaxTru: We could do that, but then we would introduce arrow functions (which we don't have yet). Therefore I stepped away from it.

But what I will do is bind the context via `apply` to get away from the `self = this`

```
    var open = wrapForCompatibility(function open(view) {
      return this._switchView(view);
    });

    return open.apply(this, arguments);
```

packages/dmn-js-shared/src/util/CompatabilityUtils.js
```diff
@@ -0,0 +1,56 @@
+import { isFunction } from 'min-dash';
```
barmac, line 1: There's a typo in the file name :)

```diff
- CompatabilityUtils
+ CompatibilityUtils
```

packages/dmn-js/test/spec/NavigatedViewerSpec.js
MaxTru: Yep - I plan to migrate to `async` tests once I have the key APIs promisfied. I think I would have to add many comments, so I would propose to not do that.

I already put a comment in the commit, that migrating away from `done` in the tests is planned as part of this promisification endavour.

packages/dmn-js-shared/src/util/CompatabilityUtils.js
MaxTru: Ups, thank you

MaxTru: Fixed

packages/dmn-js-shared/src/base/Manager.js
barmac: How about this?

```diff
diff --git a/packages/dmn-js-shared/src/base/Manager.js b/packages/dmn-js-shared/src/base/Manager.js
index 79bc8641..4c1eb260 100644
--- a/packages/dmn-js-shared/src/base/Manager.js
+++ b/packages/dmn-js-shared/src/base/Manager.js
@@ -63,6 +63,9 @@ export default class Manager {
     this._viewers = {};
 
     this._init(options);
+
+    // keep support for callbacks
+    this.open = wrapForCompatibility(this.open.bind(this));
   }
 
   /**
@@ -347,11 +350,7 @@ export default class Manager {
    * or rejects with {OpenError}
    */
   open(view) {
-    var open = wrapForCompatibility(function open(view) {
-      return this._switchView(view);
-    });
-
-    return open.apply(this, arguments);
+    return this._switchView(view);
   }
 
   _setDefinitions(definitions) {
```

packages/dmn-js-shared/test/spec/base/ManagerSpec.js
MaxTru: Yes - the reason is that I wanted to have an easy way to test the #open API. With the existing views that we have, a thorough test is difficult, since I found it very difficult to create all the situations one could potentially encounter (e.g., having an error but also having parsingWarnings).

Using these dummy views, one can very easily mock an of such situations.

packages/dmn-js-shared/src/base/Manager.js
barmac: So we could just wrap it in the constructor.

packages/dmn-js-shared/test/spec/base/ManagerSpec.js
MaxTru: Ups, good point.

Fixed

packages/dmn-js-shared/test/spec/base/ManagerSpec.js
MaxTru: This was a bug in the test suite if I am not mistaken.

I realized it as part of a clean-up when ensuring `done` is called when using callbacks. 
Previously the `done` was not included, hence the test would never had fail.

packages/dmn-js-shared/src/base/Manager.js
MaxTru: I like @barmac 's proposal. This allows us to keep the logic unchanged, and then to have a seperate section where we do the compatibility wrapping.

I will adjust accordingly.

packages/dmn-js-shared/test/spec/base/ManagerSpec.js
MaxTru: :+1: Fixed

packages/dmn-js-shared/src/base/Manager.js
pinussilvestrus: We also use this pattern in [other places](https://github.com/camunda/camunda-modeler/blob/develop/client/src/app/App.js#L138).

packages/dmn-js-shared/test/spec/base/ManagerSpec.js
pinussilvestrus: Cool stuff :+1:

packages/dmn-js-shared/test/spec/base/ManagerSpec.js
pinussilvestrus: Nice!

packages/dmn-js-shared/src/base/Manager.js
```diff
@@ -395,7 +423,8 @@ export default class Manager {
       }) || this._getInitialView(newViews);
 
       if (!newActiveView) {
-        return this._switchView(null);
+        this._switchView(null);
+        return;
```
barmac, line 427: Let's not forget for the `importXML` migration that this schedules a new Promise, but since we don't return it, one cannot really `await` it.

My suggestion: Don't change anything right now -> tide this up in the `importXML` PR.

packages/dmn-js-drd/src/Viewer.js
MaxTru: Thanks, fixed.

I also added respective test coverage (at least improved it), but in this area it is again difficult since it is really hard to find cases where you have an error + warnings (so the error needs to be covered, so that the parsing first finds warnings, before then blowing up with an error)

barmac (approved):

PR #638, opened 2021-06-14
feat(dmn-table): focus on row added via context menu

marstamm:
related to https://github.com/camunda/camunda-modeler/issues/2259


barmac:
How about if we select a cell in the same column as previously selected?

https://user-images.githubusercontent.com/28307541/121885183-26449180-cd14-11eb-892f-ed02a0142b49.mov

So if I select cell in the second column and then add a new row, a cell in the second column of the new row should be selected.

marstamm:
Good catch, I wasn't even aware that you can open this Menu from within the cell :smile:

barmac (approved):

PR #640, opened 2021-06-21
Correct import xml event lifecycle

MaxTru:
I spotted this missing testcoverage and inconsistency in the event lifecycle and wanted to clean it up prior to finishing the dmn-js promisification. This allows me to have a stable baseline for the promisification exercise.

### DIFF
--- a/packages/dmn-js-shared/src/base/Manager.js
+++ b/packages/dmn-js-shared/src/base/Manager.js
@@ -109,7 +109,7 @@ export default class Manager {
       }
 
       if (err || !options.open) {
-        this._emit('import.done', { error: err, warmings: parseWarnings });
+        this._emit('import.done', { error: err, warnings: parseWarnings });
 
         return done(err, parseWarnings);
       }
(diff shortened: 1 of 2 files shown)


packages/dmn-js-shared/src/base/Manager.js
```diff
@@ -109,15 +109,20 @@ export default class Manager {
...
+        this._emit('import.done', { error: err, warnings: parseWarnings });
 
         return done(err, parseWarnings);
       }
 
       var view = this._activeView || this._getInitialView(this._views);
 
       if (!view) {
-        return done(new Error('no displayable contents'));
+        var noDisplayableContentsErr = new Error('no displayable contents');
```
nikku, line 120: Great catch.

I believe I found a similar inconsistency in bpmn-js as we introduced our modern import :smile:.

As `var` in ES5 gets hoisted to the outer scope my proposal would be to simply override the existing `err` variable. We should avoid declaring a nested `var`.

nikku (changes requested):
Great stuff and good improvement on our existing test coverage :clap:.

It is not tested, it is likely broken :skull:

packages/dmn-js-shared/src/base/Manager.js
MaxTru: Thanks, good point. I corrected that.

nikku (approved):
:clap:

PR #643, opened 2021-06-23
fix: correct inter-package imports

MaxTru:
Correct inter-package imports (same way we already do it, for example [here](https://github.com/bpmn-io/dmn-js/blob/develop/packages/dmn-js-decision-table/src/Editor.js#L15))

### DIFF
--- a/packages/dmn-js-literal-expression/src/Viewer.js
+++ b/packages/dmn-js-literal-expression/src/Viewer.js
@@ -17,11 +17,11 @@ import TextareaModule from './features/textarea';
 import ViewDrdModule from './features/view-drd';
 
 /**
- * @typedef {import('../../dmn-js-shared/src/base/View).OpenResult} OpenResult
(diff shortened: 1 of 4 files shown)


barmac (approved):

packages/dmn-js-literal-expression/src/Viewer.js
```diff
@@ -17,11 +17,11 @@ import TextareaModule from './features/textarea';
 import ViewDrdModule from './features/view-drd';
 
 /**
- * @typedef {import('../../dmn-js-shared/src/base/View).OpenResult} OpenResult
+ * @typedef {import('dmn-js-shared/lib/base/View).OpenResult} OpenResult
  */
 
 /**
- * @typedef {import('../../dmn-js-shared/src/base/View).OpenError} OpenError
+ * @typedef {import('dmn-js-shared/lib/base/View).OpenError} OpenError
```
nikku, line 24: Good one :clap:

Issue #645, opened 2021-06-29
Post release steps fail


nikku:
__What should we do?__

Post release steps are supposed to update the bpmn.io demo. Instead these add bogus tags to dmn-js (`dmn-js-v....`): Cf. [latest post-release build](https://github.com/bpmn-io/dmn-js/runs/2896440700?check_suite_focus=true).

* [x] Fix post release steps
* [x] Ensure that a failure in the post release step gets noticed (error handling is enabled for shell or something)

__Why should we do it?__

Fix it so we we can rely on a stable build infrastructure.

PR #646, opened 2021-06-29
Improve POST_RELEASE steps

nikku:
* enable strict error handling in scripts (https://github.com/bpmn-io/dmn-js/commit/96e409c330b26808ba8110b13f47646d0dfa17b2)
* correct credentials used (https://github.com/bpmn-io/dmn-js/commit/bb0686e8ffcbf40430528225b215b4e8866b2406)

Moving to organization provided credentials, a thing we adopted for `@bpmn-io/form-js` already, too.

---

Closes https://github.com/bpmn-io/dmn-js/issues/645.

### DIFF
--- a/tasks/stages/update-demo
+++ b/tasks/stages/update-demo
@@ -12,7 +12,7 @@ CLONE_DIR="$WORKDIR/bpmn-io-demo"
 # create work dir
 mkdir -p "$WORKDIR"
 
-git clone --depth=1 "https://$GH_AUTH@github.com/$BPMN_IO_DEMO_ENDPOINT.git" "$CLONE_DIR"
+git clone --depth=1 "https://$BPMN_IO_TOKEN@github.com/$BPMN_IO_DEMO_ENDPOINT.git" "$CLONE_DIR"
 
 cd "$CLONE_DIR"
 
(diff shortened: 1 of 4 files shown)


tasks/stages/update-demo
```diff
@@ -1,5 +1,8 @@
 #! /bin/bash
 
+set -euo pipefail
```
barmac, line 3: Is it necessary? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

tasks/stages/update-demo
```diff
@@ -1,5 +1,8 @@
...
+shopt -s inherit_errexit nullglob
```
barmac, line 4: What's this for? 🤡

tasks/stages/update-demo
nikku: It is, as I can run the same script locally. In that case it should exhibit the same behavior, too.

tasks/stages/update-demo
nikku: Did some research and decided to settle down on that configuration [based on this article](https://dev.to/banks/stop-ignoring-errors-in-bash-3co5).

nikku: My goal is to catch as many errors as possible, hence that additional line (described in the article above, too).

barmac: Thanks for sharing knowledge.

barmac (approved):

PR #652, opened 2021-07-29
fix(viewer): break long words into multiple lines

### LINKED ISSUE #651: Display of Decision Tables differ from modeler to viewer
__Describe the Bug__

Decision tables with long entries get broken into multiple lines in the modeler, but overlap in the viewer

Modeler:
image

Viewer:
image

__Steps to Reproduce__

1. Open the Decision in this code sandbox: https://codesandbox.io/s/awesome-hill-pu7ds
2. the inputs overlap

Compare it to the same diagram in the modeler:
https://codesandbox.io/s/heuristic-hypatia-cz3i1

__Expected Behavior__

The viewer also uses line breaks

__Additional Info__
Caused because in the modeler, the content is wrapped in a `div` with the default browser style of `overflow-wrap: break-word;`. This is not set if the text node is child of the `td`.

Related to https://jira.camunda.com/browse/CAM-13753

__Environment__

 - Browser: Chrome 91
 - OS: Ubuntu
 - Library version: 11.0.1

marstamm:
closes #651

### DIFF
--- a/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
+++ b/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
@@ -87,6 +87,7 @@
 .dmn-decision-table-container td {
   position: relative;
   white-space: pre-wrap;
+  overflow-wrap: break-word;
   background-clip: padding-box;
 }
 


pinussilvestrus (approved):
Good stuff 👍  Makes absolute sense to keep it consistent for both Modeler & Viewer.

Now as we are defining it, we are also independent of browser defaults, right? (as mentioned in the issue description for the Modeler)

marstamm:
Yes, we are independent of browser defaults, unless they have different behaviors for the same css attribute. I had this happen in the platform docs with `word-break: break-word` (now deprecated), so you never know :smile:

PR #653, opened 2021-08-16
refactor(dmn-js-decision-table): rename class

PKief:
Since this is a default export, it is not so important what the class is called here. However, for consistency and maintainability reasons, I would suggest naming this class also as it is called in the other components. In addition, it then corresponds to the file name.


nikku (approved):
Makes sense. 

Thanks for your contribution.

PR #666, opened 2022-01-19
Add whitespace validation for output name

### LINKED ISSUE #551: Output name must not contain white spaces

Having a whitespace in an output name will lead to an error in the decision table execution.

Add a validation to the output name field and check for whitespaces. Output name must not contain whitespaces.

n/a

__Additional context__
Screenshot_20200729_112954

smbea:
Closes #551

### DIFF
--- a/packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/OutputEditor.js
+++ b/packages/dmn-js-decision-table/src/features/decision-table-head/editor/components/OutputEditor.js
@@ -1,7 +1,7 @@
 import { Component } from 'inferno';
 
 import ContentEditable from 'dmn-js-shared/lib/components/ContentEditable';
-import Input from 'dmn-js-shared/lib/components/Input';
+import ValidatedInput from 'dmn-js-shared/lib/components/ValidatedInput';
 
 
 export default class OutputEditor extends Component {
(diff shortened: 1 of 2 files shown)


pinussilvestrus:
I am just wondering, will whitespaces in the output name still crash with the new, FEEL-based execution engine? I can remember that whitespaces in variable names are okay in FEEL.

smbea:
> I am just wondering, will whitespaces in the output name still crash with the new, FEEL-based execution engine? I can remember that whitespaces in variable names are okay in FEEL.

@pinussilvestrus Uhm, if we move into Feel expressions for the output name, then this wouldn't work. For FEEL, this would be a string in quotes, right? Because if so, then we can check: (1) if in quotes -> whitespaces ok; (2) if not -> no whitespaces.

barmac:
According to DMN spec, there are no constraints regarding output name. It's type is `string` and as such can contain spaces. Cf. https://www.omg.org/spec/DMN/1.3/PDF#page=91 
The XSD for DMN 1.3 also uses `xsd:string` type without additional constraints:
```xml
	<xsd:complexType name="tOutputClause">
		<xsd:complexContent>
			<xsd:extension base="tDMNElement">
				<xsd:sequence>
					<xsd:element name="outputValues" type="tUnaryTests" minOccurs="0"/>
					<xsd:element name="defaultOutputEntry" type="tLiteralExpression" minOccurs="0"/>
				</xsd:sequence>
				<xsd:attribute name="name" type="xsd:string" use="optional"/>
				<xsd:attribute name="typeRef" type="xsd:string"/>
			</xsd:extension>
		</xsd:complexContent>
	</xsd:complexType>
```

Let's not forget that `dmn-js` is a general DMN tool. We shouldn't add Camunda-specific constraints, especially as we are working towards decoupling the Camunda Platform part from the DMN part.

MaxTru:
Sorry, being the issue opener back in 2020 that is on me.

But I agree - in particular with the "linting" approach we are working on (which could then in fact actually validate engine-specifics), validating this forecully directly in the diagram does not make much sense

smbea:
Glad we can close this issue then, even if it's because it lost relevance

PR #667, opened 2022-01-19
Select new shape after replace

smbea:
Closes camunda/camunda-modeler/issues/2306

https://user-images.githubusercontent.com/25825387/150164290-3af25fe3-8d7d-4c00-9a45-13f5aab4d11b.mov


marstamm (approved):
Works for me!

PR #671, opened 2022-01-25
Fix import in DRD

barmac:
- chore: use `no-restricted-imports` instead of import plugin
- fix(drd): fix import from dmn-js-shared

The plugin didn't catch the bug we accidentally introduced. Fortunately, there's a built-in rule which helps.


MaxTru (approved):
The no-restricted-imports is anyhow easier to configure.

Damn this happened again :-(

PR #681, opened 2022-02-16
feat(decision-table): drop `camunda:inputVariable` support

### LINKED ISSUE #680: Remove Input Variable
In FEEL, the variable can be accessed via `?`.

barmac:
This makes sure we don't use vendor extension in the decision
table. In FEEL, you can use `?` to access the input variable.

Closes #680

BREAKING CHANGES:

* The input variable field is removed. Use [camunda/camunda-dmn-js](https://github.com/camunda/camunda-dmn-js)
  to support adding and modifying `camunda:inputVariable`.


barmac:
I left this because it tests a general case of extensions and not the input variable specifically. It could as well be `foo:bar`.

marstamm:
I thought so, thanks for clarifying :+1:

PR #687, opened 2022-02-22
Display EL select only when necessary

### LINKED ISSUE #686: Do not display EL dropdown when there's only one option and the set value is the default one
__What should we do?__

Do not display EL dropdown when there's only one option and the set value is the default one.

Display here:

image

but not here:

image

__Why should we do it?__

To not confuse the user.

barmac:
EL select is displayed for two cases:
* the configured EL is not the default one
* there are multiple ELs available

* [x] decision table
* [x] literal expression

Closes #686


barmac:
@MaxTru @pinussilvestrus ready for review.

barmac:
I think we could add some documentation for config, but as a separate issue.

MaxTru (approved):

PR #695, opened 2022-04-08
fix: make sure InputSelect options are always accessible

barmac:
https://user-images.githubusercontent.com/28307541/162735393-6d90632f-593c-446c-a3df-8ab7b11e50d0.mov

Related to https://github.com/camunda/camunda-modeler/issues/2908


pinussilvestrus (approved):
LGTM 👍

PR #698, opened 2022-04-13
fix: clear state during reimport

### LINKED ISSUE #697: `importXML` breaks the instance if currently active view no longer exists
__Describe the Bug__

When calling `importXML` while a decision table is open and the new xml does not contain the same decision table, the new XML is not displayed and the "View DRD" button stops working

__Steps to Reproduce__

I have prepared a minimal reproducible test case here: https://codesandbox.io/s/dmn-js-import-bug-5wub3v?file=/src/index.js

You can use this to reproduce the behavior:

1. Observe that the DRD is correctly displayed (you might need to reload the preview on the right hand side or open https://5wub3v.csb.app/ directly)
2. Click the "Load Alternative Decision" button
3. Observe that the new DMN XML is correctly displayed
4. Reload the page so that the original DRD is displayed
5. Open the decision table by clicking the small blue icon in the corner
6. Click the "Load Alternative Decision" button
7. Nothing happens (it should instead show the DRD of the newly imported XML)
8. Clicking on the "View DRD" button also no longer shows the DRD

__Expected Behavior__

When importing new XML into an existing viewer, if the currently active view does not exist in the new XML anymore, the initial view is displayed. I assume the changes need to be made somewhere around there: https://github.com/bpmn-io/dmn-js/blob/b023ba0468cc821c0dd9725dd5f150acabd602e1/packages/dmn-js-shared/src/base/Manager.js#L149-L160

__Environment__

 - Browser: Chrome 100.0.4896.88
 - OS: Windows 10
 - Library version: 12.1.0

barmac:
Closes #697

### DIFF
--- a/packages/dmn-js-drd/test/spec/import/ImportSpec.js
+++ b/packages/dmn-js-drd/test/spec/import/ImportSpec.js
@@ -16,28 +16,41 @@ describe('DRD - Import', function() {
 
   describe('events', function() {
 
-    beforeEach(bootstrapModeler(exampleXML));
+    let events = [];
+
+    const TRACKED_EVENTS = [
+      'import.start',
+      'import.done',
+      'drdElement.added'
+    ];
+
+    beforeEach(bootstrapModeler(exampleXML, {
+      additionalModules: [{
+        __init__: [ 'eventListener' ],
+        eventListener: [ 'type', createEventListener(TRACKED_EVENTS, events) ]
+      }]
+    }));
+
+    beforeEach(function() {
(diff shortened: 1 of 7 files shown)


barmac:
So this fails on CI because we expect the view type to be kept after re-import... which is contrary to greenfield start I implemented.

barmac:
@pinussilvestrus @philippfromme @Skaiir Ready for review

philippfromme (approved):
Well done! 🎖️

PR #699, opened 2022-05-16
fix: edit button no longer always shows on scroll in decision table

Skaiir:
Closes https://github.com/bpmn-io/dmn-js/issues/668

firefox_tebMxBit0w

This was a small oversight to apply the debounced hide then show even when the button was already hidden to start with.


philippfromme (approved):
💹

PR #705, opened 2022-07-05
Model boxed expression of BKM

### LINKED ISSUE #704: I can model boxed expression of BKM
__What should we do?__

Allow to model boxed expressions for Business Knowledge Models.

__Why should we do it?__

BKM is one of the core constructs in DMN. It allows to encapsulate reusable logic. However, in the current state, the element is useless because users cannot model the logic in the tool.

barmac:
Closes #704


barmac:
Closed in favor of https://github.com/bpmn-io/dmn-js/pull/852

PR #708, opened 2022-08-05
fix(drd): fix svg creation

marstamm:
Regression from the diagram-js@8 upgrade

related to https://github.com/camunda/camunda-modeler/issues/3068


philippfromme (approved):
✅

PR #711, opened 2022-08-17
Bump dependencies

nikku:
Doing some housekeeping on our dmn-js dependencies.

Non breaking changes from what I tested, but good to keep things updated.


barmac (approved):
:rocket:

PR #721, opened 2022-12-02
add type=button to view drd button

tamas-nemeth:
the default button type is submit, so this button can cause issues inside a form


nikku (approved):
Thanks for your contribution. Makes sense.

PR #736, opened 2023-01-31
feat: break long expression names

marstamm:
Recording 2023-01-31 at 14 20 51

- show the label as `title` so hover always shows complete label
- break label up to 3 lines
- Use ellipses afterwards to prevent super long labels or expressions of taking up the whole screen

Note: I used `-webkit-line-clamp` for multi-line ellipsis. This is supported in all major browsers (except IE, cf. https://caniuse.com/css-line-clamp)

### DIFF
--- a/packages/dmn-js-decision-table/src/features/decision-table-head/components/DecisionTableHead.js
+++ b/packages/dmn-js-decision-table/src/features/decision-table-head/components/DecisionTableHead.js
@@ -143,13 +143,15 @@ function DefaultInputHeaderCell(props, context) {
 
       {
         label ? (
-          <div className="input-label" title={ translate('Input Label') }>
+          <div
+            className="input-label"
+            title={ translate('Input Label: ') + label }>
             { label }
           </div>
         ) : (
           <div
             className="input-expression"
-            title={ translate('Input Expression') }>
+            title={ translate('Input Expression: ') + inputExpression.text }>
             { inputExpression.text }
           </div>
         )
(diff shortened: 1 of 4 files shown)


packages/dmn-js-decision-table/src/features/decision-table-head/components/DecisionTableHead.js
```diff
@@ -143,7 +143,7 @@ function DefaultInputHeaderCell(props, context) {
 
       {
         label ? (
-          <div className="input-label" title={ translate('Input Label') }>
```
nikku, line 146: Shall we keep the prefix `Input Label` here? The intention would be to still tell users what this is, but also provide full details: `Input Label - SOME TEXT THAT I ENTERED AS A USER` for example.

marstamm: Makes sense, I just noticed we do it for the decision Name already
https://github.com/bpmn-io/dmn-js/blob/3081fd9284520120043697de51bfa5fdd8fb57fc/packages/dmn-js-decision-table/src/features/decision-table-properties/components/DecisionTablePropertiesComponent.js#L31

marstamm: Adjusted with [6466c76](https://github.com/bpmn-io/dmn-js/pull/736/commits/6466c7682767f8d4b595f7d3877ec15b5b156d33)

nikku (approved):

Issue #719, opened 2022-11-10
Decision table input/output label is being cut off


nikku:
__Describe the Bug__

When entering a longer decision table name the name is being cut of despite space being available vertically. Trying to hover over the name doesn't show it to me either, but shows "Input Label" instead: 

![capture iZ8ysZ_optimized](https://user-images.githubusercontent.com/58601/201037946-4a79e91c-ef49-4bd6-b232-fe6546661425.gif)

__Steps to Reproduce__

1. open [test diagram](https://github.com/bpmn-io/dmn-js/files/9978786/long-label.dmn.txt)
2. see that label is cut off and cannot be retrieved without double clicking the header

Can be reproduced on https://demo.bpmn.io/dmn (just decrease secreen size)

__Expected Behavior__

* [ ] We show the label on a best effort basis, i.e. multi-line if needed. This is a practice that other tools follow, too (cf. Excel / Google sheets)

__Environment__

Please complete the following information:

 - Browser: Any
 - OS: Any
 - Library version: 13.0.0

nikku:
@marstamm I presume this is closed via https://github.com/bpmn-io/dmn-js/pull/736?

marstamm:
Yes, closed via #736 . Missed to put the `closes` tag in the commit

GeethaParthasarathy:
This issue is reproducible for the output label and hence needs to be reopened.

<img width="960" alt="fe14a3a0d96d8c579ae06644c6986395" src="https://user-images.githubusercontent.com/85225325/217863844-cd4af352-fcd1-4882-a9de-33fc9f9b0696.png">

nikku:
Closed via https://github.com/bpmn-io/dmn-js/pull/737.

PR #737, opened 2023-02-09
fix(decision-table): clamp output label

marstamm:
Ensures that all header labels are capped at 3 lines

image

raised in slack: https://camunda.slack.com/archives/GP70M0J6M/p1675953193428329?thread_ts=1675874910.181129&cid=GP70M0J6M

----

closes https://github.com/bpmn-io/dmn-js/issues/719

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v3
     - name: Use Node.js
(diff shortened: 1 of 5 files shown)


marstamm:
closes https://github.com/bpmn-io/dmn-js/issues/719

nikku (approved):
Works for me.

One feedback (out of scope for this fix): I don't think we properly handle long labels in the editor yet: 

image

:see_no_evil:

PR #745, opened 2023-03-23
Bump rollup

nikku:
Update to rollup@3 and general house keeping.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4373,8 +4373,9 @@
     },
     "node_modules/@types/yauzl": {
       "version": "2.10.0",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -32134,7 +32135,9 @@
       "dev": true
     },
     "webpack": {
-      "version": "5.75.0",
```
smbea, line 31618: Question: why are these changes not updates in `package.json`

nikku: Because I used `npm update` over `npm install`.

Both things work :).

barmac (approved):
LGTM

PR #748, opened 2023-04-04
fix(literal-expression): avoid translate to fail with undefined typeRef

vpellegrino:
Closes https://github.com/camunda/web-modeler/issues/4252

There are situations when the `typeRef` related to a Literal Expression is just undefined. In these situations an error is thrown:
```
react_devtools_backend.js:2655 unhandled error in event listener TypeError: Cannot read properties of undefined (reading 'replace')
    at LiteralExpressionPropertiesComponent.translate [as _translate] (translate.js?e4f0:27:1)
    at LiteralExpressionPropertiesComponent.render (LiteralExpressionPropertiesComponent.js?7b38:13:415)
```
image

With the proposed change, the `translate` function would return `undefined` (instead of blocking with an error), so the hyphen will be shown for this particular "Variable Type":
image

### DIFF
--- a/packages/dmn-js-literal-expression/src/features/literal-expression-properties/components/LiteralExpressionPropertiesComponent.js
+++ b/packages/dmn-js-literal-expression/src/features/literal-expression-properties/components/LiteralExpressionPropertiesComponent.js
@@ -27,7 +27,8 @@ export default class LiteralExpressionPropertiesComponent extends Component {
           <tr>
             <td>{ this._translate('Variable Type:') }</td>
             <td>
-              <span>{ this._translate(variable.typeRef) || '-' }</span>
+              <span>{ this._translate(variable.typeRef || '') || '-' }</span>
+
             </td>
           </tr>
           <tr>


vpellegrino:
@nikku I cannot set you as a reviewer, but a reviewer from your side would be highly appreciated.

packages/dmn-js-literal-expression/src/features/literal-expression-properties/components/LiteralExpressionPropertiesComponent.js
```diff
@@ -27,7 +27,7 @@ export default class LiteralExpressionPropertiesComponent extends Component {
           <tr>
             <td>{ this._translate('Variable Type:') }</td>
             <td>
-              <span>{ this._translate(variable.typeRef) || '-' }</span>
+              <span>{ this._translate(variable.typeRef ?? '') || '-' }</span>
```
nikku, line 30:
```suggestion
              <span>{ this._translate(variable.typeRef || '') || '-' }</span>

```

Let us stick to good old JS here. Not even sure what `??` does.

Could you share an example diagram for which that one is problematic? I'd love to add it as a test case.

vpellegrino: Hi @nikku, thanks for asking. Let me give a bit more context on this.
Nullish coalescing is [JS standard](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing). It is a bit more specific than the `OR` operator (covers `null` and `undefined` cases, leaving out other falsy values).

However, `OR` is probably more aligned with the existing codebase, so I'll commit to your suggestion.

Below you can find the XML of the DMN diagram that causes the failure:
```
<?xml version="1.0" encoding="UTF-8"?>
<definitions xmlns="https://www.omg.org/spec/DMN/20191111/MODEL/" xmlns:dmndi="https://www.omg.org/spec/DMN/20191111/DMNDI/" xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/" xmlns:modeler="http://camunda.org/schema/modeler/1.0" xmlns:camunda="http://camunda.org/schema/1.0/dmn" id="Definitions_1urqfvl" name="DRD" namespace="http://camunda.org/schema/1.0/dmn" exporter="Camunda Web Modeler" exporterVersion="dev" camunda:diagramRelationId="6d44610f-09df-44e2-a801-e7b15e2460ce" modeler:executionPlatform="Camunda Cloud" modeler:executionPlatformVersion="8.1.0">
  <decision id="Decision_03j2fif" name="Decision 1">
    <variable id="InformationItem_13nz6j9" />
    <literalExpression id="LiteralExpression_1yv48vf" />
  </decision>
  <dmndi:DMNDI>
    <dmndi:DMNDiagram>
      <dmndi:DMNShape id="DMNShape_0jeimi8" dmnElementRef="Decision_03j2fif">
        <dc:Bounds height="80" width="180" x="160" y="100" />
      </dmndi:DMNShape>
    </dmndi:DMNDiagram>
  </dmndi:DMNDI>
</definitions>
```

nikku: We try to be simple over fancy, thanks for the change.

nikku:
I cannot reproduce the issue on [our demo, stock dmn-js](https://demo.bpmn.io/dmn) with the example you provided.

vpellegrino:
> I cannot reproduce the issue on [our demo, stock dmn-js](https://demo.bpmn.io/dmn) with the example you provided.

The website you provided is opening a dmn editor. Where can I access a dmn viewer instead, please?

nikku:
@vpellegrino Ah, that was the context that I was missing. Will try to reproduce against the viewer and create a test case for it.

nikku:
Thanks for your contribution. Works for me.

vpellegrino:
That's great! Thanks for covering this test case @nikku

nikku:
Fixed in `v14.1.2`.

PR #750, opened 2023-04-14
fix(decision-table): fix crash when typeRef is missing

barmac:
Related to https://github.com/camunda/camunda-modeler/issues/3553


philippfromme (approved):
✅

PR #782, opened 2023-09-25
fix: display FEEL autocomplete suggestions in correct position

barmac:
Related to https://github.com/camunda/camunda-modeler/pull/3881#issuecomment-1733249226

Learning: `transform`s break computed positions in autocomplete, so we should try to avoid it if possible.

### DIFF
--- a/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
+++ b/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
@@ -59,7 +59,6 @@
 .dmn-decision-table-container .tjs-container {
   display: flex;
   flex-direction: column;
-  position: relative;
   width: min-content;
   max-width: 100%;
   height: 100%;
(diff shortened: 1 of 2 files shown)


packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
```diff
@@ -430,13 +429,9 @@
...
 
 /* powered by */
-.dmn-js-parent {
-  /* required to apply position: fixed correctly */
-  transform: translate(0);
-}
 
 .dmn-decision-table-container .powered-by {
-  position: fixed;
+  position: absolute;
```
nikku, line 434: How does this behave with scrolling (huge tables)?

nikku: Anyhow we can also sort that out in a follow-up.

nikku (approved):

barmac:
Big table:

https://github.com/bpmn-io/dmn-js/assets/28307541/fa3b7223-de37-4ca6-b0f1-8183ab4a8fb0

PR #783, opened 2023-09-25
Increase line height in decision table cell editor

barmac:
Before:

image

After:

image


nikku (approved):
Great stuff! Looks much cleaner.

PR #802, opened 2023-11-16
Add tab focus & keyboard enter event to DrillDown & ViewDrd

### LINKED ISSUE #778: Support DMN diagram accessibility

We are working on improving our accessibility by providing keyboard support that allowed user to be able to access the diagram using tab on the keyboard. Right now we don't see any kind of accessibility support for this library

Support tab behaviour 

This could be worth checking https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard

__Additional context__

None

BrianJVarley:
> 🔎 a11y changes to improve accessibility of the expand and minimise Decision Table diagram

-  `ViewDrd`: Remove button outline style override, allow native browser focus style to be applied on focus
- `DrillDown`: Change element type to - `button` to allow keyboard focus and capture ENTER key event to trigger drilldown.

Closes #778  (💡 Discussion in progress & tests needed)

### DIFF
--- a/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
+++ b/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
@@ -426,11 +426,6 @@
   background: var(--view-drd-button-hover-background-color);
 }
 
-/* Adds the missing focus outline to button */
-/* .dmn-decision-table-container .view-drd .view-drd-button:focus {
-  border: 1px solid var(--view-drd-button-focus-border-color);
-} */
-
 /* end view drd */
 
 /* powered by */
--- a/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
+++ b/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
@@ -38,6 +38,7 @@
   --view-drd-button-border-color: var(--color-grey-225-10-75);
   --view-drd-button-color: var(--color-grey-225-10-15);
   --view-drd-button-hover-background-color: var(--color-grey-225-10-97);
+  --view-drd-button-focus-border-color: var(--color-blue-205-100-50);
   --decision-table-font-family: 'Arial', sans-serif;
 
   width: 100%;
@@ -424,6 +425,11 @@
 
 .dmn-decision-table-container .view-drd .view-drd-button:hover {
   background: var(--view-drd-button-hover-background-color);
+}
+
+/* Adds the missing focus outline to button */
+.dmn-decision-table-container .view-drd .view-drd-button:focus {
+  border: 1px solid var(--view-drd-button-focus-border-color);
 }
 
 /* end view drd */
--- a/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
+++ b/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
@@ -38,7 +38,6 @@
   --view-drd-button-border-color: var(--color-grey-225-10-75);
   --view-drd-button-color: var(--color-grey-225-10-15);
   --view-drd-button-hover-background-color: var(--color-grey-225-10-97);
-  --view-drd-button-focus-border-color: var(--color-blue-205-100-50);
   --decision-table-font-family: 'Arial', sans-serif;
 
   width: 100%;
--- a/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
+++ b/packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
@@ -420,7 +420,6 @@
   color: var(--view-drd-button-color);
   font-weight: bold;
   cursor: pointer;
-  outline: none;
 }
 
 .dmn-decision-table-container .view-drd .view-drd-button:hover {
@@ -428,9 +427,9 @@
 }
 
 /* Adds the missing focus outline to button */
-.dmn-decision-table-container .view-drd .view-drd-button:focus {
+/* .dmn-decision-table-container .view-drd .view-drd-button:focus {
   border: 1px solid var(--view-drd-button-focus-border-color);
-}
+} */
 
 /* end view drd */
 
--- a/packages/dmn-js-drd/src/features/drill-down/DrillDown.js
+++ b/packages/dmn-js-drd/src/features/drill-down/DrillDown.js
@@ -87,10 +87,6 @@ export default class DrillDown {
     // TODO(nikku): can we remove renamed to drillDown.enabled
     if (this._config.enabled !== false) {
       domClasses(html).add('interactive');
-
-      // Add tabindex attr. to allow keyboard tab focus on Drilldown button
-      domAttr(html, 'tabindex', '0');
-
       this.bindEventListener(element, html, overlayId);
     }
   }
--- a/packages/dmn-js-drd/src/features/drill-down/DrillDown.js
+++ b/packages/dmn-js-drd/src/features/drill-down/DrillDown.js
@@ -71,7 +71,7 @@ export default class DrillDown {
   addOverlay(element, className) {
     const html = domify(`
       <div class="drill-down-overlay">
-        <button type="button" class="${className}"></button>
+        <button type="button" class="${className} drill-down-btn"></button>
       </div>
     `);
 
--- a/packages/dmn-js-drd/src/features/drill-down/DrillDown.js
+++ b/packages/dmn-js-drd/src/features/drill-down/DrillDown.js
@@ -3,7 +3,6 @@ import { is } from 'dmn-js-shared/lib/util/ModelUtil';
 import {
   domify,
   classes as domClasses,
-  attr as domAttr,
   delegate as domDelegate
 } from 'min-dom';
 
--- a/packages/dmn-js-drd/src/features/drill-down/DrillDown.js
+++ b/packages/dmn-js-drd/src/features/drill-down/DrillDown.js
@@ -71,7 +71,7 @@ export default class DrillDown {
   addOverlay(element, className) {
     const html = domify(`
       <div class="drill-down-overlay">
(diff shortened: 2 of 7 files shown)


packages/dmn-js-drd/src/features/drill-down/DrillDown.js
```diff
@@ -71,7 +72,7 @@ export default class DrillDown {
   addOverlay(element, className) {
     const html = domify(`
       <div class="drill-down-overlay">
-        <span class="${className}"></span>
+        <button class="${className}"></button>
```
BrianJVarley, line 75: Since the span element doesn't capture ENTER key, I've changed this element to a Button which does capture it.
But also open to other ideas on how to bind the click event using span.

packages/dmn-js-decision-table/assets/css/dmn-js-decision-table.css
```diff
@@ -426,6 +427,11 @@
   background: var(--view-drd-button-hover-background-color);
 }
 
+/* Adds the missing focus outline to button */
+.dmn-decision-table-container .view-drd .view-drd-button:focus {
+  border: 1px solid var(--view-drd-button-focus-border-color);
+}
```
BrianJVarley, line 433: The ViewDrd button had no focus outline comparing with DrillDown button, so I've added one.
But there may be alternative solutions you are considering to add this focus style.

packages/dmn-js-drd/src/features/drill-down/DrillDown.js
```diff
@@ -87,6 +88,9 @@ export default class DrillDown {
     if (this._config.enabled !== false) {
       domClasses(html).add('interactive');
 
+      // Add tabindex attr. to allow keyboard tab focus on Drilldown button
+      domAttr(html, 'tabindex', '0');
```
BrianJVarley, line 92: Adding the tabindex=0 attribute now allows the elements to be focused via tab key.

packages/dmn-js-drd/test/spec/features/drill-down/DrillDownSpec.js
```diff
@@ -252,6 +254,28 @@ describe('features - drilldown', function() {
...
+
+        // when
+        // press ENTER
+        triggerFocusIn(overlayEl);
+        triggerKeyEvent(overlayEl, 'keydown', 13);
+
+        // then
+        expect(drillSpy).to.have.been.calledOnce;
+
+        expect(drillSpy).to.have.been.calledWith(overlayEl);
```
BrianJVarley, line 275: Working on fixing this test, at the moment overlayEl does return the button element, but the drillSpy is not called in test.

packages/dmn-js-drd/src/features/drill-down/DrillDown.js
BrianJVarley: This is not removed in latest commit, since the button element is being included in tab focus.

barmac:
Hi, thanks for opening this PR and sorry that you had to wait for so long. I will look into this today.

packages/dmn-js-drd/src/features/drill-down/DrillDown.js
barmac: We still need to fix styling here. Also, if the drilldown is not enabled (e.g. for DRD-only viewer), this should remain to be a `span` as it's not interactive.

barmac: I added `type=button` because otherwise it's `type=submit` which is not correct.

packages/dmn-js-drd/src/features/drill-down/DrillDown.js
```diff
@@ -71,7 +71,7 @@ export default class DrillDown {
   addOverlay(element, className) {
     const html = domify(`
       <div class="drill-down-overlay">
-        <span class="${className}"></span>
+        <button type="button" class="${className}"></button>
```
barmac, line 74: The styling is not quite there yet. I'd rather keep the old look for now.

BrianJVarley: @barmac can you explain a bit more about the styling? I guess you mean that the previous blue background is missing on the new button element.

**Before:**
Screenshot 2023-11-28 at 15 20 35


**After:**

Screenshot 2023-11-28 at 15 19 50

packages/dmn-js-drd/src/features/drill-down/DrillDown.js
BrianJVarley: Yeah good point, about the styling added a comment here - https://github.com/bpmn-io/dmn-js/pull/802#discussion_r1407841798

I guess we want to add a blue background for the drill down button? 
Similar to what we had with the span element.

BrianJVarley: > Also, if the drilldown is not enabled (e.g. for DRD-only viewer), this should remain to be a span as it's not interactive.

Is there some state we can check for "DRD-only viewer" mode? Then we could just add a conditional to render either a span or button:

```
let html;
    if (isDrdOnlyMode) {
      html = domify(`
      <div class="drill-down-overlay">
        <span type="button" class="${className}"></span>
      </div>
    `);
    } else {
      html = domify(`
      <div class="drill-down-overlay">
        <button type="button" class="${className}"></button>
      </div>
    `);
    }
```

packages/dmn-js-drd/src/features/drill-down/DrillDown.js
BrianJVarley: I've pushed a new commit to match the styling that was there previously on the span.
What do you think? __https://github.com/bpmn-io/dmn-js/pull/802/commits/34041ad1349528a740d1e369514069831a915292

_Without focus_

Screenshot 2023-11-28 at 15 42 03

_With focus_

Screenshot 2023-11-28 at 15 41 50

barmac: What I meant is that if we switch to standard button look, we change the visuals significantly. On the other hand, we need **some** way to indicate focus.

BrianJVarley: @barmac I've pushed [another commit](https://github.com/bpmn-io/dmn-js/pull/802/commits/e3d7e1484280e0d21e32ddf1061c4d9a8722cd6b) so the focus style is more apparent on the drill down button. 

Do you see some specific style difference between original and new style as described below?
If so I can try to adjust it further to match the expectation.


For comparison this is the _original span style on focus_, i.e, no focus focus style applied just span with blue background:

Screenshot 2023-12-04 at 11 27 24

Whereas, this is the latest commit blue button which IMO looks the same as original span style, _but with the browser focus style being applied_:


_Before focus (default style)_


Screenshot 2023-12-04 at 11 24 34


_On focus (Dark blue focus style applied by browser)_

Screenshot 2023-12-04 at 11 24 40

barmac: Thanks, it looks better now. I will tweak it a bit and then merge this PR :)

BrianJVarley: Great, looking forward to the next release then :)

barmac:
I just noticed that we set the icon DOM element to be a button even if it's not clickable. I will fix that.

barmac:
OK I played a bit with CSS, and also made sure that we display `span` and not `button` if the drilldown icon is not interactive. Before we merge this, I'd love to gather feedback from the bpmn.io team.

I built a demo website which can be accessed at https://dmn-js-a11y-demo.netlify.app/

https://github.com/bpmn-io/dmn-js/assets/28307541/f1fa3ea9-a40f-45c4-83b8-8d6e3fad52bf

marstamm:
[X-Posting from Slack for transparency]

The Focus has little contrast on the DRD, I think. For the first iteration, this is absolutely fine (still better than what we had before), but we could consider changing it in a follow up

I had a look how Github does it for dark buttons, and they add a white border around the blue outline. Not sure if this is something we want to do
image (5)
image (6)

barmac:
I tested this additionally on Firefox, and the focus there is even harder to notice.

What we need to remember is that this PR is just a first step. We can do much more for the accessibility of the tools. For example, the a11y tools don't announce the element which is focused. However, I think it's still better to have some improvement than no-solution as it is without this contribution.

lmbateman:
Thank you for raising the accessibility issue, @marstamm . I think there are three ways we could address it:

1. If we have control over the focus color, we could darken it. Then it would contrast well with both the white background and the button color.
2. We could also try lightening the button color instead, just on focus. But if we take this route, I think we need to consider it more holistically.
3. Add an outline-offset to the focus as Martin suggested. This might be difficult because of the lack of space and the small size of the button, but I think reducing the button background area by a pixel on each side would work from a visual perspective (not sure if it's feasible from a technical perspective).

Otherwise, this looks good to me.

barmac:
Thanks @BrianJVarley for initiating this and everybody else for feedback! :)

BrianJVarley:
> [X-Posting from Slack for transparency]
> 
> The Focus has little contrast on the DRD, I think. For the first iteration, this is absolutely fine (still better than what we had before), but we could consider changing it in a follow up
> 
> I had a look how Github does it for dark buttons, and they add a white border around the blue outline. Not sure if this is something we want to do image (5) image (6)

@marstamm @barmac  I've noticed also using screen reader that the view / hide DRD buttons don't have an [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label). I think it would improve the accesibility to label these buttons also for example "Show DRD" / "Hide Diagram"

Screenshot 2023-12-11 at 10 57 53

barmac:
Thanks for the feedback. Indeed, right now the user does not hear what the button action is. This could be something like _View decision table/literal expression of {name}_.

PR #818, opened 2024-01-17
chore(context-pad): user friendly titles

smbea:
Similar to https://github.com/bpmn-io/bpmn-js/commit/aa158e2d5a7de86b8a388945bbb56f217922f1ae


nikku (approved):

nikku:
@smbea Let's release this and ensure we fix `bpmn-io-i18n`, too.

PR #832, opened 2024-02-28
chore: transpile to supported browsers only

barmac:

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -3,7 +3,14 @@
     [
       "@babel/preset-env", {
         "modules": false,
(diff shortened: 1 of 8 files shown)


nikku:
Love this. Actually prepared it myself some time ago, but did not find the time to fully execute it :tada:.

barmac:
Impact:

Before:
```
du -sh lib dist
 32K	lib
6.5M	dist
```

After:
```
du -sh lib dist
 32K	lib
6.1M	dist
```

nikku:
Incredible!

Issue #839, opened 2024-03-22
Fix post release job


nikku:
### What should we do?

The post release job does not actually do a wait (loop) but [fails immediately](https://github.com/bpmn-io/dmn-js/actions/runs/8394273637/job/22991017264). We have to fix this.

### Why should we do it?

Ensure that our infrastructure is reliable.

---

Related to https://github.com/bpmn-io/bpmn-js/issues/2058. Let's take this as an opportunity to once again offer a solid baseline:

* operate on tags
* properly await publish in registry

PR #840, opened 2024-03-22
fix(POST_RELEASE): properly await publishing

nikku:
Closes https://github.com/bpmn-io/dmn-js/issues/839


barmac (approved):
Let's try this out.

barmac:
It seems to still not work correctly: https://github.com/bpmn-io/dmn-js/actions/runs/8568362978

PR #843, opened 2024-04-03
Fix issues found with axe

### LINKED ISSUE #821: DMN elements - Scrollable region must have keyboard access
### Describe the Bug

The hit policy selector and the DMN table are affected by accessibility issue: **Scrollable region must have keyboard access**.

AXE recommendation:
To solve this problem, you need to fix at least (1) of the following:
* Element should have focusable content
* Element should be focusable

image

### Steps to Reproduce

1. Open the DMN editor (for example on the Web Modeler)
2. Open the hit policy selector
3. Resize the page so that the internal scrollbar appears on both elements (see screenshot)
4. Execute the page scan, by using the ["axe DevTools" extension](https://www.deque.com/axe/devtools/)

### Expected Behavior

No accessibility issue is reported by the axe scan.

### Environment

 - Browser: Chrome
 - OS: macOs
 - Library version: current one (reproducible on https://demo.bpmn.io/dmn/new)

barmac:
Related to https://github.com/camunda/product-hub/issues/2189

Closes #821

### DIFF
--- a/packages/dmn-js-decision-table/src/features/decision-rules/components/DecisionRulesCellEditorComponent.js
+++ b/packages/dmn-js-decision-table/src/features/decision-rules/components/DecisionRulesCellEditorComponent.js
@@ -130,9 +130,7 @@ class TableCellEditor extends Component {
   isDefaultExpressionLanguage(businessObject) {
     const { expressionLanguage } = businessObject;
 
-    const defaultExpressionLanguage = this.getDefaultExpressionLanguage(
-      businessObject
-    ).value;
+    const defaultExpressionLanguage = this.getDefaultExpressionLanguage().value;
 
     return !expressionLanguage || expressionLanguage === defaultExpressionLanguage;
   }
@@ -144,18 +142,18 @@ class TableCellEditor extends Component {
   getExpressionLanguageLabel(businessObject) {
     const { expressionLanguage } = businessObject;
 
-    const defaultExpressionLanguage = this.getDefaultExpressionLanguage(businessObject);
+    const defaultExpressionLanguage = this.getDefaultExpressionLanguage();
 
     return this._expressionLanguages.getLabel(expressionLanguage) ||
       defaultExpressionLanguage.label;
   }
 
-  isScript(businessObject) {
-    const defaultExpressionLanguage = this.getDefaultExpressionLanguage(businessObject);
+  isScript() {
+    const { businessObject } = this.props;
 
-    const isInputCell = is(businessObject, 'dmn:UnaryTests');
+    const defaultExpressionLanguage = this.getDefaultExpressionLanguage();
 
-    if (!isInputCell) {
+    if (!this._isInputCell()) {
       return false;
     }
 
@@ -167,8 +165,12 @@ class TableCellEditor extends Component {
       businessObject.expressionLanguage !== defaultExpressionLanguage;
   }
 
-  getDefaultExpressionLanguage(businessObject) {
-    const elementType = is(businessObject, 'dmn:UnaryTests') ? 'inputCell' : 'outputCell';
+  _isInputCell() {
+    return is(this.props.businessObject, 'dmn:UnaryTests');
+  }
+
+  getDefaultExpressionLanguage() {
+    const elementType = this._isInputCell() ? 'inputCell' : 'outputCell';
 
     return this._expressionLanguages.getDefault(elementType);
   }
@@ -185,7 +187,7 @@ class TableCellEditor extends Component {
     const { businessObject } = this.props;
 
     return businessObject.expressionLanguage ||
-      this.getDefaultExpressionLanguage(businessObject).value;
+      this.getDefaultExpressionLanguage().value;
   }
 
   _getVariables() {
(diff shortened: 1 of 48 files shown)


barmac:
I will push more changes to this PR so that we can merge it more easily. Otherwise I'd have to open PRs based on this branch.

packages/dmn-js-decision-table/src/features/decision-rules/components/DecisionRulesCellEditorComponent.js
```diff
@@ -167,8 +163,12 @@ class TableCellEditor extends Component {
...
   }
 
-  getDefaultExpressionLanguage(businessObject) {
-    const elementType = is(businessObject, 'dmn:UnaryTests') ? 'inputCell' : 'outputCell';
+  _isInputCell() {
+    return is(this.props.businessObject, 'dmn:UnaryTests');
+  }
+
+  getDefaultExpressionLanguage() {
+    const elementType = this._isInputCell() ? 'inputCell' : 'outputCell';
```
marstamm, line 171: As I understand this, This is just a refactoring to get the businessObject from the props instead of having it passed to the function, correct?

If so, we can then also adjust the function call in `isScript` for consistency. (non-blocking, can be done as follow-up)

https://github.com/bpmn-io/dmn-js/blob/49a586d6a6e5abf548b65afc14bde2163ab998ae/packages/dmn-js-decision-table/src/features/decision-rules/components/DecisionRulesCellEditorComponent.js#L152

packages/dmn-js-drd/src/features/drill-down/DrillDown.js
```diff
@@ -93,6 +91,28 @@ export default class DrillDown {
...
+
+    const button = document.createElement('button');
+    button.type = 'button';
+    button.className = className;
+    button.title = this._translate(title);
+
+    container.appendChild(button);
+
+    return container;
+  }
```
marstamm, line 114: What is the reason we are not using `domify` here? Personally, I find it easier to read than building with native API.

The function name can be confusing, as we return a DOM node and not HTML. We could consider renaming it to `_getOverlayNode`, but we do use it interchangeably in other places

packages/dmn-js-drd/test/spec/features/drill-down/DrillDownSpec.js
```diff
@@ -88,6 +88,37 @@ describe('features - drilldown', function() {
 
       });
 
+
+      describe('accessibility', function() {
+
+
+        verify(
```
marstamm, line 95: Code Style
```suggestion
      describe('accessibility', function() {

        verify(
```

packages/dmn-js-shared/test/spec/components/EditableComponentSpec.js
```diff
@@ -91,6 +91,19 @@ describe('components/EditableComponent', function() {
...
+      <TestComponent value={ null } label="label" />
+    );
+
+    // then
+    expect(node.innerHTML).to.match(/aria-label="label"/);
+  });
+
+
+
   describe('empty class', function() {
```
marstamm, line 107: Code Style
```suggestion
  });


  describe('empty class', function() {
```

packages/dmn-js-decision-table/src/features/decision-rules/components/DecisionRulesCellEditorComponent.js
barmac: OK

packages/dmn-js-drd/src/features/drill-down/DrillDown.js
barmac: I wanted to reduce the risks of having accidental XSS vector. Note that `domify` doesn't care whether the result of `translate` call contains quotation marks which otherwise would have to be escaped.

barmac: I renamed the method.

packages/dmn-js-shared/test/spec/components/EditableComponentSpec.js
barmac: If only there was a linting rule for this 🤡

marstamm:
I found 1 more issue with the axe dev tool. The Expression language dropdown does not have a label. I am not sure if this is relevant at all, as we do not have this field in C8.
image

Otherwise, looks very clean! Great work 👏

packages/dmn-js-drd/src/features/drill-down/DrillDown.js
marstamm: Makes sense, let's keep it like this then 👍

marstamm (approved):
🚀

barmac:
> fix: ensure accessible names for expression language select

Done via https://github.com/bpmn-io/dmn-js/pull/843/commits/af13b8dcef9642bcdd00c38ee800a762363e184b

packages/dmn-js-shared/test/spec/components/EditableComponentSpec.js
nikku: It is [you to create one](https://github.com/bpmn-io/eslint-plugin-bpmn-io) :wink:.

PR #844, opened 2024-04-04
test: setup accessibility tests with axe

barmac:
This will ensure that we don't violate basic principles in the initial view.


barmac:
Right now the test fails due to empty heading for the index column. In DMN, this heading is reserved for the hit policy which we moved to the top a couple of years ago. Hit policy is unrelated to the index though.

barmac:
From the spec:

image

packages/dmn-js/test/spec/ModelerSpec.js
```diff
@@ -1,6 +1,6 @@
 import Modeler from 'src/Modeler';
 
-import { insertCSS } from 'test/helper';
+import { expectToBeAccessible, insertCSS } from 'test/helper';
```
nikku, line 3: :star:

nikku (approved):

PR #851, opened 2024-04-18
feat(label-editing): hide labels while editing

marstamm:
related to https://github.com/bpmn-io/diagram-js-direct-editing/issues/23

This PR ensures we hide the labels on the diagram when direct editing is active. This aligns the [behavior with bpmn-js](https://github.com/bpmn-io/bpmn-js/blob/07f75f7cd5d2ef9934d7b32cccfee6d4f05b8fc6/lib/features/label-editing/LabelEditingPreview.js#L76-L84) and enables us to remove the white background in https://github.com/bpmn-io/diagram-js-direct-editing/issues/23

### DIFF
--- a/packages/dmn-js-drd/src/features/label-editing/LabelEditingProvider.js
+++ b/packages/dmn-js-drd/src/features/label-editing/LabelEditingProvider.js
@@ -9,6 +9,7 @@ import {
   isDefined
 } from 'min-dash';
 
+var MARKER_LABEL_HIDDEN = 'djs-label-hidden';
 
 export default function LabelEditingProvider(
     canvas,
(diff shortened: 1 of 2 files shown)


nikku:
@marstamm Could you attach a screen capture of the proposed behavior?

marstamm:
For now, nothing will change as the labels are currently "hidden" by the white background of direct editing. With https://github.com/bpmn-io/diagram-js-direct-editing/pull/39 , we will not hide drilldown/non-square shapes by the white background:

Recording 2024-04-18 at 14 24 05

Without this change, the original label would still be visible while editing:
Recording 2024-04-18 at 14 25 29

nikku (approved):
Makes sense, thanks!

PR #868, opened 2024-05-23
fix: make add row button focussable

### LINKED ISSUE #846: Cannot focus "add row" button
### Describe the Bug

The "add row" button (+) is inaccessible with the keyboard as it's a `span` without `tabindex`.

### Steps to Reproduce

Try to focus it.

### Expected Behavior

It should be a `button` with accessible label.

### Environment

 - OS: [e.g. Windows 7]

---

Child of https://github.com/camunda/product-hub/issues/2189

barmac:
https://github.com/bpmn-io/dmn-js/assets/28307541/152f5e0f-f4b7-4d86-8573-e79d37ae221d

Closes #846


nikku (approved):
I verified that this is needed. `TAB` expectedly shall move the focus, including this element; `ENTER` will also create a new row (common in table editors).

Implementation looks solid and simple enough, too. :clap:

PR #869, opened 2024-05-23
fix: make add column buttons focussable

### LINKED ISSUE #847: Cannot focus "add column" button
### Describe the Bug

The "add column" button (+) is inaccessible with the keyboard as it's a span without tabindex.

### Steps to Reproduce

Try to focus it.

### Expected Behavior

It should be a button with accessible label.

### Environment

 - OS: [e.g. Windows 7]

---

Child of https://github.com/bpmn-io/internal-docs/issues/909

barmac:
Closes #847


nikku (changes requested):
While it is technically focusable now I'd consider to fine-tune the styling a little bit:

capture bGVPMP_optimized

One way to do it to add an `outline-offset: 4px;`, resulting in this:

capture GTD6YX_optimized

barmac:
Indeed, the outline looks invisible on your screencast. In my browser (Chrome on Mac), it is easily distinguishable.

barmac:
I've added the change from your suggestion.

barmac:
Thanks for checking.

image

nikku (approved):
Works great :cake:

PR #874, opened 2024-05-27
test: use spy to check if variables were retrieved

### LINKED ISSUE #873: Fix flaky test in literal expression
### What should we do?

Fix the test failing randomly, e.g. https://github.com/bpmn-io/dmn-js/actions/runs/9251123564/job/25445979185

### Why should we do it?

Trust in CI.

barmac:
We can't reliably ensure that the autocompletion opens in tests, so this solution makes the test use a mock variable resolver instead.

Closes #873

### DIFF
--- a/packages/dmn-js-literal-expression/test/spec/features/textarea/TextareaEditorSpec.js
+++ b/packages/dmn-js-literal-expression/test/spec/features/textarea/TextareaEditorSpec.js
@@ -1,8 +1,8 @@
+/* global sinon */
+
 import { bootstrapModeler, inject } from 'test/helper';
 
 import { query as domQuery } from 'min-dom';
-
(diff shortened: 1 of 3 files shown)


packages/dmn-js-literal-expression/test/spec/features/textarea/TextareaEditorSpec.js
```diff
@@ -90,26 +96,23 @@ describe('textarea editor', function() {
...
-        const options = testContainer.querySelectorAll('[role="option"]');
-
-        expect(options).to.exist;
-        expect(options).to.satisfy(options => {
-          const result = Array.from(options).some(
-            option => option.textContent === 'Variable');
-          return result;
-        });
-      });
+      expect(getVariablesSpy).to.have.been.called;
```
nikku, line 115: Much better, more robust test :clap:

nikku (approved):

Issue #880, opened 2024-06-03
Parse decision table inputs as unary tests


barmac:
### Describe the Bug

Currently, the decision table parses input cells values as expressions. This leads to misleading syntax errors as in the screenshot below:

![image](https://github.com/bpmn-io/dmn-js/assets/28307541/e64c7b17-23cf-48f9-b7f2-6899b73a7d60)

### Steps to Reproduce

Pass unary tests separated by comma to the input cell.

### Expected Behavior

Should be parsed as Unary Test.

### Context

Reported via https://camunda.slack.com/archives/C0693F1NFK5/p1717418704931359

Ideally, this should be configurable in the [feel-editor](https://github.com/bpmn-io/feel-editor/tree/main)

nikku:
Upstream fix pending integration: https://github.com/bpmn-io/feel-editor/pull/53.

barmac:
Reported again https://camunda.slack.com/archives/C0693F1NFK5/p1731061838985009

PR #881, opened 2024-06-04
feat(decision-table): parse FEEL input cells as unary tests

nikku:
Closes https://github.com/bpmn-io/dmn-js/issues/880


barmac:
Implemented via https://github.com/bpmn-io/dmn-js/pull/947

Issue #848, opened 2024-04-05
Cannot use context menu with keyboard


barmac:
### Describe the Bug

The context menu does not take the focus and I cannot take any action in it with keyboard.

![image](https://github.com/bpmn-io/dmn-js/assets/28307541/2d2324c5-4be5-48bb-b9ba-b85b4cc606ac)

How to open context menu with keyboard on Mac (thanks @lmbateman): https://support.apple.com/en-gb/guide/mac-help/mh27469/mac

### Steps to Reproduce

1. Open context menu.
2. Try to move around with arrow keys.

### Expected Behavior

Context menu should work exactly like a system context menu:

* [ ] Be a focus trap
* [ ] Change selection with arrow keys
* [ ] Select and perform action with enter

It closes with escape or outside click already, which otherwise would have been another requirement.

### Environment

Please complete the following information:

 - Browser: [e.g. IE 11, Chrome 69]
 - OS: [e.g. Windows 7]
 - Library version: [e.g. 2.0.0]

barmac:
Note that this issue is not detected by axe.

barmac:
I am not working on this right now. As this is not essential for axe-compliance, I am moving this to the backlog.

PR #891, opened 2024-08-06
feat: decision table context menu keyboard navigation

jarekdanielak:
### Proposed Changes

Closes https://github.com/bpmn-io/dmn-js/issues/848

Context menu support for keyboard navigation.

https://github.com/user-attachments/assets/f98c6c00-b610-4fc1-9432-8b85260e0998

 

### Checklist

* [X] **Brief textual description** of the changes present
* [X] **Visual demo** attached
* [X] Related issue linked via `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`

### DIFF
--- a/packages/dmn-js-decision-table/src/features/context-menu/ContextMenuKeyboard.js
+++ b/packages/dmn-js-decision-table/src/features/context-menu/ContextMenuKeyboard.js
@@ -1,30 +1,32 @@
-import { query, classes } from 'min-dom';
+import { query, queryAll, classes } from 'min-dom';
 
 export default class ContextMenuKeyboard {
 
-  DIRECTION_UP = -1;
-  DIRECTION_DOWN = 1;
-
   constructor(eventBus) {
-    eventBus.on('contextMenu.open', () => {
-      document.addEventListener('keydown', this.handleKeyEvent);
-      document.addEventListener('mouseover', this.handleMouseOver);
-    });
+    eventBus.on('contextMenu.open', () => this.addEventListeners());
 
-    eventBus.on('contextMenu.close', () => {
-      document.removeEventListener('keydown', this.handleKeyEvent);
-      document.removeEventListener('mouseover', this.handleMouseOver);
-    });
+    eventBus.on('contextMenu.close', () => this.removeEventListeners());
+    eventBus.on('commandStack.executed', () => this.removeEventListeners());
   }
+
+  addEventListeners = () => {
+    document.addEventListener('keydown', this.handleKeyEvent);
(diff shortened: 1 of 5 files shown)


barmac:
While the key navigation works fine, Voice Over does not announce the active element. We need some ARIA attributes to make it work.

Screenshot 2024-08-06 at 17 07 31

barmac:
This could be helpful: https://www.w3.org/WAI/ARIA/apg/patterns/menubar/
I'd consider to implement focus wrapping, i.e. arrow down on the last item should focus the first item.

barmac:
@philippfromme perhaps you could have a look independently? So that the solutions meets the expectations of two different platforms users.

barmac:
Technically, we target features at `develop`, and `fix`es at `main`.

barmac:
I don't quite get it why capslock+left/right from voiceover does not move the focus, but I guess we have to live with it. I think this PR is an improvement good enough to merge.

lerna.json
```diff
@@ -1,4 +1,4 @@
 {
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
-  "version": "16.6.1"
+  "version": "16.6.2"
```
barmac, line 3: Make sure to rebase so that the git history is clean.

barmac (changes requested):

barmac:
There's a merge conflict again 😓

lerna.json
barmac: The "publish" commit is still on the branch. Shall we look together into rebase?

jarekdanielak: Rabased 🥳

jarekdanielak:
Commit history is now clean and there are no merge conflicts.

If @philippfromme verifies it on Windows, we can marge.

packages/dmn-js-decision-table/src/features/context-menu/ContextMenuKeyboard.js
```diff
@@ -0,0 +1,132 @@
...
+
+    hover && disableHoverStyle(hover);
+    classes(current).remove('focused');
+    classes(next).add('focused');
+
+    // If mouse is currently over the entry that we are moving to...
+    next.style.removeProperty('background-color');
+  };
+
+  open = () => {
```
philippfromme, line 114: Can we rename this method? `open` is misleading as this just clicks the selected entry which results in closing the context menu.

philippfromme (commented):
Consider renaming `open`. Also, let's squash the commits before merging. Otherwise looking good and works on Windows, too. ✅

PR #909, opened 2024-10-15
Prevent global `keydown` event when `ContentEditable` handled `ENTER`

nikku:
### Proposed Changes

This fixes a regression introduced via a6dfc16b9cd6ebf56da14b532ccca728550cb0d8.

The prior change switched to `onKeyDown`, which is an synthetic event handler that cannot prevent default (on global clicks).

What we need to prevent default (global) clicks is actual DOM event handling. Quoting from the Inferno documentation [1]:

> Lower case events will bypass Inferno's event system in favour of
> using the native event system supplied by the browser.

[1]: https://www.infernojs.org/docs/guides/event-handling

---

The fixed behavior can be inspected via `npm start`, using the Dish Decision `JavaScript` field, but also in the fact that the test suite passes (again):

capture he9DAr_optimized

### Checklist

* [x] **Brief textual description** of the changes present
* [x] **Visual demo** attached
* [x] Related issue linked via `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`

### DIFF
--- a/packages/dmn-js-shared/src/components/ContentEditable.js
+++ b/packages/dmn-js-shared/src/components/ContentEditable.js
@@ -243,11 +243,11 @@ export default class ContentEditable extends Component {
         spellCheck="false"
         data-placeholder={ placeholder || '' }
         onInput={ this.onInput }
-        onkeypress={ this.onkeypress }
+        onkeypress={ this.onkeypress } // intentionally lowercase to use native event
         onPaste={ this.onPaste }
         onFocus={ this.onFocus }
         onBlur={ this.onBlur }
-        onkeydown={ this.onkeydown }
+        onkeydown={ this.onkeydown } // intentionally lowercase to use native event
         ref={ node => this.node = node }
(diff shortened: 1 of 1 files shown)


nikku:
Rebased on top of [fixed](https://github.com/bpmn-io/dmn-js/commit/f71de90e35b9853e95706ff05c55b624c1951d6a) `develop`. Let's not squash things without adjusting to a conventional commit.

CC @jarekdanielak.

barmac:
I find it super confusing that Inferno supports both lowercase and camelcase event handlers. Anyway, thanks for fixing this!

barmac:
I've added comments which should save us some WTF in the future.

barmac (approved):

barmac:
I've prepared a PR aiming to fix the flaky test causing CI to fail: https://github.com/bpmn-io/dmn-js/pull/910

nikku:
This still works as expected. The comments are a valuable addition :heavy_check_mark:

PR #917, opened 2024-11-08
fix: make name change behavior not break on name change

barmac:
### Proposed Changes

https://github.com/user-attachments/assets/73f5fbd9-1ead-4cbb-81fd-8c2dcc6644c1

We've tested the label update but not the direct name change with `modeling.updateProperties`. Now we test for both.

Related to https://github.com/bpmn-io/dmn-js-properties-panel/issues/103 
Related to https://github.com/camunda/camunda-modeler/issues/4684

### Checklist

* [x] **Brief textual description** of the changes present
* [x] **Visual demo** attached
* [x] Related issue linked via `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`

### DIFF
--- a/packages/dmn-js-drd/test/spec/features/modeling/behavior/NameChangeBehaviorSpec.js
+++ b/packages/dmn-js-drd/test/spec/features/modeling/behavior/NameChangeBehaviorSpec.js
@@ -6,7 +6,7 @@ import CoreModule from 'src/core';
 import Modeling from 'src/features/modeling';
 
 
-describe('NameChangeBehavior', function() {
+describe('features/modeling - NameChangeBehavior', function() {
 
   beforeEach(bootstrapModeler(simpleStringEditXML, {
     modules: [
--- a/packages/dmn-js-drd/test/spec/features/modeling/behavior/NameChangeBehaviorSpec.js
+++ b/packages/dmn-js-drd/test/spec/features/modeling/behavior/NameChangeBehaviorSpec.js
@@ -8,66 +8,117 @@ import Modeling from 'src/features/modeling';
 
 describe('NameChangeBehavior', function() {
 
-  describe('with label change', function() {
-
-    beforeEach(bootstrapModeler(simpleStringEditXML, {
-      modules: [
-        CoreModule,
-        Modeling
-      ],
-    }));
-
-    describe('should update variable name when label is changed', function() {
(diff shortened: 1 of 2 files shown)


barmac:
After a second look at the name change behavior, I am not super happy about having two components doing the same thing (boxed expression and shared). However, for this fix I decided to only make a small change to avoid a potential bug.

barmac:
Once merged, this will be released as 16.8.2

packages/dmn-js-drd/test/spec/features/modeling/behavior/NameChangeBehaviorSpec.js
```diff
@@ -8,66 +8,117 @@ import Modeling from 'src/features/modeling';
 
 describe('NameChangeBehavior', function() {
```
nikku, line 9:
```suggestion
describe('features/modeling - NameChangeBehavior', function() {
```

nikku (approved):
We missed updating the `name` attribute via modeling APIs and I see this is now covered.

Relevant follow-up: Will it break if we update moddle properties?

barmac:
But I'd do it in a separate PR. Ignoring an update is not breaking the editor.

barmac:
Published as v16.8.2

barmac:
Hmm this job should not have run at all: https://github.com/bpmn-io/dmn-js/actions/runs/11739770879
We need to fix it later on.

nikku:
I think it would be reasonable to check for `@latest` tag and if that does not exist (because a release is a backport) the job is fine to fail.

By the way, your release broke `latest`, it is now back to `16.x`, cf. [npm](https://www.npmjs.com/package/dmn-js?activeTab=versions).

barmac:
Thanks for notice! I'll fix that

barmac:
This will be fixed via v17.0.1 release.

barmac:
Hmm I can't release it because tests fail due to f402e252332aa53bc8e23c2872716e521db7c197. I will have a look again after lunch.

barmac:
As a temporary fix, I fixed the `latest` for dmn-js. The subpackages will be fixed with the v17.0.1

barmac:
So the test is now broken although its `expect`s are logical. I debugged the test case, and while we set incorrect classes in the code, the actual position of the simple mode button is correct (top-right but with `.bottom.right` classes).

Issue #945, opened 2025-05-06
Include named exports of the Viewer, NavigatedViewer & Modeler classes


BrianJVarley:
### Describe the Bug

In Angular applications & libs, some development teams ban internal path imports from npm libraries. ([no-restricted-imports](https://typescript-eslint.io/rules/no-restricted-imports/). In dmn-js lib there is no named exports for viewers only default export, which results in an import path as - `import Viewer from 'dmn-js/lib/NavigatedViewer`. [This gist **](https://gist.github.com/everget/52041712ec5a14627031d553a2510cf4)also** outlines some of the cons of using only default exports.

I'm proposing not so much a fix but just additionally exporting the viewer and modeler classes as named exports for convenience. 

### Steps to Reproduce

1. Import NavigatedViewer using named import - `import NavigatedViewer from 'dmn-js';`
2. Import is undefined, only default import exists with internal path - `import Viewer from 'dmn-js/lib/NavigatedViewer`

### Expected Behavior

The library provides a named export for each viewer and modeler class. This is generally the norm for library imports within the Angular framework. 

### Environment

Please complete the following information:

 - N/A
 - N/A
 - Library version: `17.2.0`

barmac:
Hi,

Thanks for opening the issue. While I'm not necessarily convinced that disallowing nested paths in imports is a good choice, I see some value in using named exports from the root as you suggested. This could simplify the usage of the library.
Still, since we'd rather roll out such change across the core bpmn.io libraries (bpmn-js, dmn-js, form-js), we need to discuss it internally first. I will keep you posted.

BrianJVarley:
@barmac yeah that makes sense, agree it would simplify the usage of the library for devs.

PR #946, opened 2025-05-06
dmn-js: Viewers index.js: Use Viewer named exports to improve 

BrianJVarley:
### Proposed Changes

Related to [{I945}](https://github.com/bpmn-io/dmn-js/issues/945)

- This changes includes the Viewer, NavigatedViewer & Modeler as named exports from the dmn-js library.
Exporting these classes instead of just default has some benefits for consumers. We don't need to use the long import path `import Viewer from 'dmn-js/lib/NavigatedViewer'` instead `import { NavigatedViewer } from 'dmn-js';`. It also means you don't need to re-alias the Viewer when importing.

**For example with named exports:**

```js
export { default } from './Viewer';
export { Viewer } from './Viewer';
export { NavigatedViewer } from './NavigatedViewer';
export { Modeler } from './Modeler';
```

### Steps to reproduce the behavior:

Import NavigatedViewer using named import - `import { NavigatedViewer } from 'dmn-js`';
Import is undefined, only default import exists with internal path - `import Viewer from 'dmn-js/lib/NavigatedViewer`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "lerna": "^8.1.9",
         "mocha": "^10.8.2",
         "mocha-test-container-support": "^0.2.0",
-        "npm-run-all2": "^7.0.1",
+        "npm-run-all2": "^8.0.0",
         "puppeteer": "^24.0.0",
         "sinon": "^17.0.1",
         "sinon-chai": "^3.7.0",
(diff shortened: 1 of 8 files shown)


barmac:
Cf. https://github.com/bpmn-io/dmn-js/issues/945#issuecomment-2866121267

BrianJVarley:
Feel free to close this, as you mentioned it needs a discussion.

PR #991, opened 2026-03-30
fix: fix error on keyboard navigation in popups without menu entries

AlekseyManetov:
Related to https://github.com/camunda/camunda-modeler/issues/4958

### Proposed Changes

Fix error on keyboard navigation in popups without menu entries

https://github.com/user-attachments/assets/0dc46c95-86f1-485f-a7d4-1af36ae00713

### Checklist

Ensure you provide everything we need to review your contribution:

* [x] Contribution __meets our [definition of done](https://github.com/bpmn-io/.github/blob/main/resources/DEFINITION_OF_DONE.md)__
* [x] Pull request __establishes context__
  * [x] __Link to related issue(s)__, i.e. `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`
  * [x] __Brief textual description__ of the changes
  * [x] __Screenshots or short videos__ showing UI/UX changes
  * [ ] __Steps to try out__, i.e. [using the `@bpmn-io/sr` tool](https://github.com/bpmn-io/sr)

### DIFF
--- a/packages/dmn-js-decision-table/test/spec/features/context-menu/ContextMenuKeyboardSpec.js
+++ b/packages/dmn-js-decision-table/test/spec/features/context-menu/ContextMenuKeyboardSpec.js
@@ -17,6 +17,8 @@ import CoreModule from 'src/core';
 import InteractionEventsModule from 'table-js/lib/features/interaction-events';
 import ModelingModule from 'src/features/modeling';
 import DecisionRulesModule from 'src/features/decision-rules';
+import DecisionTableHeadModule from 'src/features/decision-table-head';
+import DecisionTableHeadEditorModule from 'src/features/decision-table-head/editor';
 
 describe('context menu keyboard', function() {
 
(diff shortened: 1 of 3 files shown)


packages/dmn-js-decision-table/test/spec/features/context-menu/ContextMenuKeyboardSpec.js
```diff
@@ -119,6 +119,22 @@ describe('context menu keyboard', function() {
...
+
+    // given
+    // simulate a popup without navigable entries (e.g. input expression editor)
+    eventBus.fire('contextMenu.open', {
+      position: { x: 0, y: 0 },
+      context: { contextMenuType: 'input-edit' }
+    });
+
+    // when / then
+    expect(() => {
```
ev-codes, line 140: **suggestion:** consider asserting that:
- the context menu is actually open
- the context menu has keyboard focus

This will help ensure the test is recreating the conditions in the initial bug issue.

ev-codes (approved):
Thanks for this!

I left one comment, but otherwise LGTM!

packages/dmn-js-decision-table/test/spec/features/context-menu/ContextMenuKeyboardSpec.js
AlekseyManetov: Thanks for pointing out, i've a bit reworked the test to ensure that we recreating the bug conditions.

> The context menu has keyboard focus

This doesn't make sense, since `ContextMenuKeyboard` listens on document, not on the `.context-menu element`, so we catch any keydown if menu is open.

